Issue #31, opened 2021-05-05
[Bug/Typo] The :pg_version: entity is not expanding in Installation sections (was: Windows exe link broken)


pfried:
**Add the appropriate label(s) -->**

**Describe the bug**

The Link for downloading the windows version of timescale does not work, the variable substitution seems broken:

`https://timescalereleases.blob.core.windows.net/windows/timescaledb-postgresql-:pg_version:_x.y.z-windows-amd64.zip`

**Location**

https://docs.timescale.com/timescaledb/latest/how-to-guides/install-timescaledb/self-hosted/windows/installation-windows/

**Screenshots**

**Device and browser**
Desktop

Loquacity:
Thanks for reporting this. I'm having trouble reproducing the issue, the link looks fine to me on my browser. We just refreshed the docs, perhaps there's a caching issue? Can you take another look for me please?

![image](https://user-images.githubusercontent.com/3914967/117223397-e2c35180-ae50-11eb-801b-48b71085da1b.png)

pfried:
![image](https://user-images.githubusercontent.com/1744875/117244302-17c3aa00-ae39-11eb-81c5-41fcc5fbe4b3.png)

yes the issue is gone, but you still seem to have problems with interpolation of placeholders, see marking in your screenshot (same for my view)

Loquacity:
You are absolutely correct, the `:pg_version:` entity looks broken throughout. Thanks!

PR #35, opened 2021-05-07
Remove pg_version entity and replace with version.

Loquacity:
The `:pg_version:` entity was not being replaced in the Installation docs. After discussion with @tedsczelecki we decided to hard code it. I've gone with 13, as it's the latest version, LMK if it should be 12.

Fixes: https://github.com/timescale/docs/issues/31


jonatas (approved):
Great work! and thank you for removing all spaces at the end of the lines 👯‍♀️

jonatas:
I think we can point the base to `master` instead of `latest`.  @tedsczelecki confirm?

ryanbooz (approved):

Loquacity:
> Great work! and thank you for removing all spaces at the end of the lines 👯‍♀️

haha, that's my editor doing the hard work there, so I can't take the credit for that 😉

PR #45, opened 2021-05-11
Add missing link, and remove admon markup

Loquacity:
# Description

Link was written, but no URL provided. This PR adds a URL, and removes the unnecessary admon markup.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Community report: https://timescaledb.slack.com/archives/C4GT3N90X/p1620772714211200

### DIFF
--- a/timescaledb/how-to-guides/continuous-aggregates/manually-refresh-specific-ranges.md
+++ b/timescaledb/how-to-guides/continuous-aggregates/manually-refresh-specific-ranges.md
@@ -15,7 +15,6 @@ would trigger a refresh of time buckets that are not yet completely
 filled. It might also materialize a lot of data, increase write
 amplification, and affect other policies such as data retention.
 
-<highlight type="tip">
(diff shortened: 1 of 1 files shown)


jonatas (commented):
I don't get why we're removing the highlight block.

AFAIR to use highlight with links we should have default MD markup for links like [anchor](URL).

Loquacity:
> I don't get why we're removing the highlight block.
> 
> AFAIR to use highlight with links we should have default MD markup for links like [anchor](URL).

Oh, I removed it more because it was a bit unnecessary. It has a lot more content in it than an admon should have,. and there's nothing in it that really needs to be called out. Additionally, it's really important to remember that readers quite often skip over admons, especially if they're over used. We need to make sure we're keeping admons to things that are actually really important, and use them sparingly.

jonatas:
Thank you @Loquacity!

jonatas (approved):

PR #58, opened 2021-05-17
FAQs

Loquacity:
# Description

Move FAQs to top level, and split

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

https://docs.google.com/document/d/1k0BVzu3gbNAj69dK4NVWHr2M7qSZyh_uA2VCLoXRRZM/edit?usp=sharing


Loquacity:
@tedsczelecki Can you tell me if I'm doing this correctly? Preview is still 403 for me :(

ryanbooz (changes requested):
Structure-wise, I'm not sure I completely understand how this gets moved up to a "product"-level alongside TimecaleDB, API, etc. If we do, however, then there are some changes you'll need to make to the Page Index file to allow it to work as you're intending.

I do think @CoolAssPuppy should way in a little. I love breaking these down, and you can better explain the potential search win we get by doing this @Loquacity

Loquacity:
> Structure-wise, I'm not sure I completely understand how this gets moved up to a "product"-level alongside TimecaleDB, API, etc. If we do, however, then there are some changes you'll need to make to the Page Index file to allow it to work as you're intending.
> 

Yeah, I'm increasingly thinking that leaving it under TimescaleDB might be easiest for now, and then we can re-think it when we implement the IA review recommendations. I'll update this PR to reflect that.

> I do think @CoolAssPuppy should way in a little. I love breaking these down, and you can better explain the potential search win we get by doing this @Loquacity

I had a quick chat with @CoolAssPuppy about this last week, but for the record:

FAQs don't serve a good purpose as documentation, because if they have useful information in them, they sit outside of the rest of the docs, meaning readers have to go hunting for the info. Additionally, having them as one long file makes them hard to read, most readers won't stick around to read all of them on the chance that they'll get an answer to a question they have. So, in most tech docs situations, I would recommend not having FAQs at all. Where FAQs *can* be useful, though, is as an SEO mechanism. To make this work, we split them into smaller files, use metadata to mark them up appropriately, and ensure that the questions we address are questions that people are putting into search engines. If we do all that correctly, then Google will show the answers to our questions as 'featured snippets' at the top of search results. So, this PR is about beginning that process by breaking them down appropriately, in preparation for editing the questions and adding that metadata.

Loquacity:
image

ryanbooz:
I'm game for that. Let's see where it leads!! :-D

ryanbooz (approved):

Issue #60, opened 2021-05-19
[Docs RFC] High-level copy-edit of all How To pages


Loquacity:
# Describe change in content, appearance, or functionality

The current How To pages (https://docs.timescale.com/timescaledb/latest/how-to-guides/) lack consistency. This issue tracks a high level review of these docs to bring them up to standard in terms of wording, writing style, syntax, etc. In preparation for a more comprehensive review of the structure later on.

# Subject matter expert (SME)

@Loquacity for writing guidance

charislam:
Finished!

PR #62, opened 2021-05-19
How-To edits: Backup/restore

Loquacity:
# Description

High level edit of the backup/restore How Tos.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Partial: https://github.com/timescale/docs/issues/60

# Notes for reviewers

Please review this and let me know what you think. It implements a topic structure that matches my IA review proposal, and sets the scene for us to include the HowTo metadata at the index level. If you're happy with this structure, I'll go ahead and do the rest of them, but let's get this one right first so we have a model.


Loquacity:
Can I please get a review or two on this? Need to know if you're happy with this structure before I go ahead and do all the others.

timescaledb/how-to-guides/backup-and-restore/docker-and-wale.md
```diff
@@ -1,206 +1,194 @@
...
-cron jobs that can invoke base backups via, e.g., the WAL-E
-container's HTTP API.
+To enable the WAL-E docker image to perform archiving, it needs to use the same
+network and data volumes as the TimescaleDB container. It also needs to know the
+location of the write-ahead log and data directories. You can pass all this
+information to WAL-E when you start it. In this example, the WAL-E image listens
+for commands on the `timescaledb-net` internal network at port 80, and writes
+backups to `~/backups` on the Docker host.
+
+### Running the WAL-E sidecar
```
ryanbooz, line 65: So, I realize this follows the above example heading-wise, but I wonder if this is supposed to start with "Procedure" as well?

timescaledb/how-to-guides/backup-and-restore/pg-dump-and-restore.md
```diff
@@ -1,114 +1,103 @@
...
-```bash
-psql -d new_db -c "SELECT create_hypertable('conditions', 'time')"
-```
-
-<highlight type="tip">
-The parameters to `create_hypertable` do not need to be
-the same as in the old db, so this is a good way to re-organize
-your hypertables (e.g., change partitioning key, number of
-partitions, chunk interval sizes, etc.).
+Do not use the `pg_dump` command to backup individual hypertables. Dumps created using this method lack the necessary information to correctly restore the hypertable from backup.
```
ryanbooz, line 20: This simplifies the content on the left quite a bit, which isn't bad, but it does feel like we're leaving something out here. In essence, the previous content was saying, "if you backup and restore TimescaleDB with `pg_dump`, it must always be restored to a PostgreSQL database running the **same** version of TimescaleDB before upgrading to a newer release of the extension."

Is it worth adding something like this in here?

timescaledb/how-to-guides/backup-and-restore/pg-dump-and-restore.md
```diff
@@ -1,114 +1,103 @@
-# Using `pg_dump`  and  `pg_restore` [](pg_dump-pg_restore)
+# Logical backups with `pg_dump`  and  `pg_restore` [](pg_dump-pg_restore)
 
-In this section, we cover how to backup and restore an entire
-database or individual hypertables using the native PostgreSQL
-[`pg_dump`][pg_dump] and [`pg_restore`][pg_restore] commands.
+You can ackup and restore an entire database or individual hypertables using the
```
ryanbooz, line 3:
```suggestion
You can backup and restore an [entire database](#anchor) or [individual hypertables](#anchor) using the
```

timescaledb/how-to-guides/backup-and-restore/pg-dump-and-restore.md
```diff
@@ -1,114 +1,103 @@
...
+    ```
+1.  Restore the database:
+    ```sql
+    \! pg_restore -Fc -d exampledb exampledb.bak
+    SELECT timescaledb_post_restore();
+    ```
+
+### Troubleshoot version mismatches
+
+The PostgreSQL `pg_dump` command does not allow you to specify which version of
```
ryanbooz, line 42: These two paragraphs needs some work to reduce duplication and redundancy. It does feel, to me, that these warnings are important enough that we should address them upfront. I mean, in theory folks aren't doing a backup and then ignoring the rest of the article... but people are lazy and it would be easy to miss this important point in my opinion.

At some point we also need to document David's backup tool, which has been on my plate for a long time. ☹️ 

https://github.com/timescale/timescaledb-backup

timescaledb/how-to-guides/backup-and-restore/pg-dump-and-restore.md
```diff
@@ -1,114 +1,103 @@
...
+the extension to use when backing up. This can create problems if you have a
+more recent version installed. For example, if you create the backup using an
+older version of TimescaleDB, and when you restore it uses the current version,
+without giving you an opportunity to upgrade first.
+
+You can work around this problem when you are restoring from backup, if you restore to a restore to a PostgreSQL instance with the same extension version as the original database before you perform the restore. After the data is restored, you can upgrade the version.
```
ryanbooz, line 48:
```suggestion
You can work around this problem when you are restoring from backup by making sure the new PostgreSQL instance has the same extension version as the original database before you perform the restore. After the data is restored, you can upgrade the version of TimescaleDB.
```

timescaledb/how-to-guides/backup-and-restore/docker-and-wale.md
```diff
@@ -1,206 +1,194 @@
...
-
-Our [WAL-E Docker image][wale image] runs a small Web endpoint that
-accepts WAL-E commands via a HTTP API. This allows PostgreSQL to
-communicate with the WAL-E sidecar over the internal network to
-trigger archiving. It is, of course, also possible to use the
-container to invoke WAL-E directly. The Docker image accepts the
-standard WAL-E environment variables to configure the archiving
-backend (e.g., AWS S3) and more. See [WAL-E's documentation][wale
+# Ongoing physical backups with Docker & WAL-E
+When you using TimescaleDB in a containerized environment, you can use
```
ryanbooz, line 2:
```suggestion
When you run TimescaleDB in a containerized environment, you can use
```

ryanbooz (changes requested):

timescaledb/how-to-guides/backup-and-restore/docker-and-wale.md
ryanbooz, line 65: I realize this isn't changing much content (if any) below, but reading over this it's not clear to me that this section will actually perform the backup. WAL-E is certainly an advanced topic so maybe we expect more experience (and if so should state somewhere), but if I'm coming to these pages looking for backup how-tos, I'd expect a "Backup" section to complement the "Recovery" section. It's just not clear to me that "running the wal-e sidecar" is going to make my backup.

Does that make sense?

Loquacity: There's no procedure here. That requires numbered steps.

Loquacity: Yep, agreed. Updated to "Perform/ing the backup using the WAL-E sidecar".

timescaledb/how-to-guides/backup-and-restore/pg-dump-and-restore.md
Loquacity: I say that in more detail under the Troubleshooting heading. Adding that info here seems like we'd be conflating two different ideas. Here in the admon, we have the idea of backing up individual hypertables, where later on we deal with the idea of version mismatches.

timescaledb/how-to-guides/backup-and-restore/pg-dump-and-restore.md
Loquacity: > These two paragraphs needs some work to reduce duplication and redundancy. It does feel, to me, that these warnings are important enough that we should address them upfront. I mean, in theory folks aren't doing a backup and then ignoring the rest of the article... but people are lazy and it would be easy to miss this important point in my opinion.

I can see the point, but surely you would only need this info if you ran into an error with the backup/restore? In which case, you'd come looking for troubleshooting info?

> 
> At some point we also need to document David's backup tool, which has been on my plate for a long time. ☹️
> 
> https://github.com/timescale/timescaledb-backup

New issue: https://github.com/timescale/docs/issues/97

timescaledb/how-to-guides/backup-and-restore/pg-dump-and-restore.md
ryanbooz: Maybe my overall point is that separating these two actions means that somone could theoretically read the section on backing up, do that for some period of time without restoring (bad practice, I know... but...) and then when they are in trouble and try to restore, it doesn't work and they're left in sticky situation. In truth, it's not the biggest, biggest deal, but _maybe_ if we said upfront that, "hey, you need to always keep track of what Postgres and TimescaleDB version you're on in case you need to restore to the exact same version..." - folks would be more inclined to keep track fo that?

I mean, in theory it's a moot point with David's tool and `pg_dump` isn't/shouldn't be your main backup mechanism, but that's where my thinking is at with it.

Loquacity: OK, that's a fair point. I'll have a play with it and see what I can come up with.

Loquacity:
I think is ready when everyone else is happy 👍

PR #124, opened 2021-06-10
How To Edits: Configuration

Loquacity:
# Description

High level edit of the configuration How Tos.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Partial: https://github.com/timescale/docs/issues/60

# Notes for reviewers

Please review this and let me know what you think. It implements a topic structure that matches my IA review proposal, and sets the scene for us to include the HowTo metadata at the index level.


Loquacity:
image

jacobprall (commented):
I'll wait for Ryan to review, but there are some redirects that I'll need to update whenever we're ready to merge

ryanbooz:
Need to add the new `about-configuration` page to the `page-index` in order to review. That said, I'm slightly conflicted by what should go in the "about" pages. I'm a big fan of the section landing pages and like how this works out in "Compression" (for instance), but here I'm not 100% sure what we've gained by taking some (but not all) things from a separate page about PG configuration and moving it into "about". To me, it feels like consistency across features would be more like:

- Landing (index) page with quick intro and bulleted "here's all of the help/how-to content we provide on this topic"
- About ({feature}-about) page that gives a little more depth on the topic as necessary. (what are the various ways you can configure the system, what does compression really mean, how does data retention work) with appropriate links
- How-to specific documents for each action

To that end, I would expect the "configuration-about" page to talk about the various tools and ways (timescaledb-tune, postgres.conf, command line), the pros/cons and then direct users to the specific page.

Maybe I'm totally wrong here and would love more discussion (which I absolutely should have started 2 months ago on this PR... 100% my fault).

Thoughts?

Loquacity:
> Need to add the new `about-configuration` page to the `page-index` in order to review. That said, I'm slightly conflicted by what should go in the "about" pages. I'm a big fan of the section landing pages and like how this works out in "Compression" (for instance), but here I'm not 100% sure what we've gained by taking some (but not all) things from a separate page about PG configuration and moving it into "about". To me, it feels like consistency across features would be more like:
> 
> * Landing (index) page with quick intro and bulleted "here's all of the help/how-to content we provide on this topic"
> * About ({feature}-about) page that gives a little more depth on the topic as necessary. (what are the various ways you can configure the system, what does compression really mean, how does data retention work) with appropriate links
> * How-to specific documents for each action
> 
> To that end, I would expect the "configuration-about" page to talk about the various tools and ways (timescaledb-tune, postgres.conf, command line), the pros/cons and then direct users to the specific page.
> 
> Maybe I'm totally wrong here and would love more discussion (which I absolutely should have started 2 months ago on this PR... 100% my fault).
> 
> Thoughts?

Ah, I see what you're saying (and thanks for adding it into the index). I was thinking that the concept in the about page was directed more to "what are the various configuration parameters I might want to adjust", rather than "what are the different ways I can change configuration parameters". Mostly because I didn't think the differences between the different methods were very remarkable. I feel like the choice between the methods becomes obvious once you know which parameter you want to adjust.

All that said, even if we move the discussion of the parameters to a different section ("Common configuration parameters" or something?), I definitely like having them all together, and using a consistent style.

Loquacity:
@ryanbooz bump?

Loquacity:
> @ryanbooz bump?

@ryanbooz bumping again now ABL is done

jacobprall (approved):

PR #151, opened 2021-07-01
Update Windows link to 2.3.0 and show both PG 12 and 13

kohenkatz:
# Description

The download link for Windows points to a version that was not the most recent one.

Also, even though PostgreSQL 12 is supported, there was no download link shown.

This PR updates the version and shows links for PG 12 and 13.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

### DIFF
--- a/timescaledb/how-to-guides/install-timescaledb/installation-windows.md
+++ b/timescaledb/how-to-guides/install-timescaledb/installation-windows.md
@@ -5,13 +5,13 @@
 #### Prerequisites
 
 - [Visual C++ Redistributable for Visual Studio 2015][c_plus] (included in VS 2015 and later)
-- A standard **PostgreSQL 13 64-bit** installation
+- A standard **PostgreSQL 12 or 13 64-bit** installation
 - Make sure all relevant binaries are in your PATH: (use [pg_config][])
 - Installation must be performed from an account with admin privileges
 
 #### Build and install
 
-1. Download the the [.zip file for your PostgreSQL version][windows-dl].
+1. Download the the .zip file for your PostgreSQL version - [12][windows-dl-12] or [13][windows-dl-13].
 1. Extract the zip file locally
 1. Run `setup.exe`, making sure that PostgreSQL is not currently running
 1. If successful, a `cmd.exe` window will pop open and you will see the following:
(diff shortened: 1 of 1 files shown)


kohenkatz:
The build link posted by GitHub Actions does not seem to actually include these changes. Why not?

Loquacity:
> The build link posted by GitHub Actions does not seem to actually include these changes. Why not?

Probably because you're working on a fork. @tedsczelecki ?

kohenkatz:
> > The build link posted by GitHub Actions does not seem to actually include these changes. Why not?
> 
> Probably because you're working on a fork.

That makes sense. I'd suggest that the GitHub Action shouldn't add links for that, but I don't know enough about GH Actions to suggest how to change it.

Loquacity:
> > > The build link posted by GitHub Actions does not seem to actually include these changes. Why not?
> > 
> > 
> > Probably because you're working on a fork.
> 
> That makes sense. I'd suggest that the GitHub Action shouldn't add links for that, but I don't know enough about GH Actions to suggest how to change it.

That's a good idea, and probably something @tedsczelecki can do. All I know is that GH actions rarely work as expected on forks.

ryanbooz:
Thanks for the contribution @kohenkatz!

There's a chance that the link doesn't work for you because it's tied to permissions of the other private repo that you don't have access to (which helps us do the build). While it is unfortunate, the link is very helpful for us, as project maintainers, so we'll keep adding it. When there is some time, we can try to dig further.

ryanbooz (approved):

Loquacity (approved):

kohenkatz:
@ryanbooz I looked into it and it seems the reason the review link does not do anything is because the other job that actually runs the PR deployment (`.github/workflows/deploy.yml`) only runs on branches in the local repository and not on forks.

The simplest thing to do is probably to add an `if` to the `pr-creation` action (based on [this](https://github.community/t/stop-github-actions-running-on-a-fork/17965/2)) so it will only run on PRs that are backed by a local branch and not a fork branch:

```
if: github.repository == 'timescale/docs'
```

Loquacity:
Thanks for looking into that @kohenkatz ! I did something similar on a GH action in a private project a while ago, so I think you're probably on the right track.

PR #163, opened 2021-07-08
Hyperfunction api reorg

ryanbooz:
# Description

This is the basic outline of changes I am proposing after digesting our conversations, reading the docs, and trying to think (ever so slightly) as a newer developer trying to understand the value of what these functions bring, etc.

The prose for the "landing" pages is mostly lifted from docs, but I'm certainly not committed to all of it. The basic flow I think this section needs to take is (effectively):
 * You're here looking at functions about percentile approximation. Cool, let's tell you generally what they are to make sure we're on the same page. Oh, and BTW, to get any value out of these, it's a two-step process. Create an aggregate and then extract the value you want from that aggregate
 * Now that you know you need an aggregate, here are the three functions you can choose from. We recommend the first one on most cases, but you can play with the others if your use case merits it.
 * So, you have an aggregate... great! Now, look at the other functions that help you extract meaningful value from that cool aggregate you now possess.  ;-) 
 
 
 I'm not saying this is a flawless pathway (or sorts), but hopefully it gets us a bit closer.
 
 **A few other notes**
 * You'll notice that the `uddsketch` and `tdigest` API pages have explanation first and then usage examples. I don't like that since it's different from everything else, but I do think some explanation is good somewhere. Thoughts?
 * I also created separate rollup documents for each input type rather than linking to one document with both in it to avoid confusion (David noticed this right away and suggested a change similar to this). We can modify as you see fit.
 * I also want to change all references of `ts` for the timestamp column to `time` since all other examples throughout documentation use that. The consistency will help IMO.


Loquacity:
LGTM, if David is happy :)

Loquacity (approved):

api/percentile-aggregation-methods.md
```diff
@@ -0,0 +1,68 @@
+# Advanced Percentile Aggregation
```
Loquacity, line 1: This page will need a copy edit, but it can probably be done after the fact.

Loquacity, line 1:
```suggestion
# Advanced percentile aggregation
```

api/percentile-aggregation-methods.md
```diff
@@ -0,0 +1,68 @@
...
+giving it greater accuracy near the tails (i.e. 0.001 or 0.995).
+- [UddSketch](/hyperfunctions/percentile-approximation/aggregation-methods/uddsketch/) – T
+his algorithm uses exponentially sized buckets to guarantee the approximation 
+falls within a known error range, relative to the true discrete percentile.
+
+The `UddSketch` algorithm underlies the `percentile_agg()` interface, it offers 
+tunability for the size and maximum error target of the sketch, while `percentile_agg` 
+uses preset defaults. 
+
+### Choosing the Right Algorithm for Your Use-Case
```
Loquacity, line 19:
```suggestion
### Choosing the right algorithm for your use case
```

api/tdigest.md
```diff
@@ -1,2 +1,56 @@
...
+| `buckets` | `INTEGER` | Number of buckets in the digest.  Increasing this will provide more accurate quantile estimates, but will require more memory.|
+| `value` | `DOUBLE PRECISION` |  Column to aggregate.
+
+### Returns
+
+|Column|Type|Description|
+|---|---|---|
+| `tdigest` | `TDigest` | A t-digest object which may be passed to other t-digest APIs. |
+
+### Sample Usages
```
Loquacity, line 43:
```suggestion
### Sample usage
```

api/time_weight.md
```diff
@@ -49,3 +48,81 @@ SELECT
     average(tw) -- extract the average from the time weight summary
 FROM t;
 ```
+
+## Advanced Usage Notes
```
Loquacity, line 52:
```suggestion
## Advanced usage notes
```

api/time_weight.md
```diff
@@ -49,3 +48,81 @@ SELECT
...
+In the LOCF approach, the value is treated as if it remains constant until the 
+next value is seen. The LOCF approach is commonly used when the sensor or 
+measurement device sends measurement only when there is a change in value.
+
+The linear interpolation approach treats the values between any two measurements 
+as if they lie on the line connecting the two measurements. The linear 
+interpolation approach is used to account for irregularly sampled data where the 
+sensor doesn't provide any guarantees.
+
+### Parallelism and Ordering
```
Loquacity, line 71:
```suggestion
### Parallelism and ordering
```

api/uddsketch.md
```diff
@@ -1,2 +1,90 @@
-# uddsketch
-The functions related to `uddsketch`.
+## uddsketch()
+Timescale's UddSketch implementation is provided as an aggregate function in 
+PostgreSQL. The output is currently only suitable as input to the 
+the percentile approximation functions. This can be directly as part of a one-off 
+SQL query, or as transient data stored in a Continuous Aggregate that is queried 
+later with these functions and using the UddSketch data as input.
+
+## Implementation Details
```
Loquacity, line 8:
```suggestion
## Implementation details
```

api/uddsketch.md
```diff
@@ -1,2 +1,90 @@
...
+it is impossible to do so while with the bucket bound.  If the sketch has had to 
+combine buckets, the new error can be found with the [uddsketch_error](#error) 
+command.
+
+Note that since the error will be increased automatically (roughly doubling at 
+each step) as the number of buckets is exceeded, it is probably worth erring on 
+the side of too small unless you have a good understanding of exactly what your 
+error should be.
+
+### Required Arguments
```
Loquacity, line 60:
```suggestion
### Required arguments
```

api/uddsketch.md
```diff
@@ -1,2 +1,90 @@
...
+
+
+### Returns
+
+|Column|Type|Description|
+|---|---|---|
+| `uddsketch` | `UddSketch` | A UddSketch object which may be passed to other UddSketch APIs. |
+
+
+### Sample Usages
```
Loquacity, line 75:
```suggestion
### Sample usage
```

Loquacity (changes requested):

PR #165, opened 2021-07-09
Add small note about timescale dependent objects

jonatas:
# Description

Today @sgichohi reached me because my instance couldn't be automatically upgraded because of the view we have in this tutorial.

We get an error like this during the upgrade:

```sql
tsdb=> ALTER EXTENSION "timescaledb" UPDATE TO "2.3.0" ;
ERROR:  cannot drop table _timescaledb_catalog.chunk because other objects depend on it
DETAIL:  view chunks_with_compression depends on table _timescaledb_catalog.chunk
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
```

So, we agreed on adding a small warning about such dependency closer to the view.

# Version

- [x] Latest (Default)


sgichohi (approved):
Awesome

timescaledb/tutorials/custom-timescaledb-dashboards.md
```diff
@@ -166,6 +166,11 @@ FROM (
     AND ch.chunk_name = cds.chunk_name;
 ```
 
+<highlight type="warning">
+Note the view is dependent on TimescaleDB internals and maybe will need to be
+dropped to upgrade the TimescaleDB extension and recreated after the upgrade.
+</highlight>
```
Loquacity, line 172:
```suggestion
<highlight type="warning">
The view is dependent on TimescaleDB internals. You might need to drop the view
to upgrade the TimescaleDB extension, and recreate it after the upgrade.
</highlight>
```

Loquacity (changes requested):
Just a wording suggestion.

Loquacity (approved):

PR #170, opened 2021-07-12
Adding very limited install instructions for launch to be revised later.

ryanbooz:
# Description

Adding a quick and dirty install section that will get further updates as more functions are released and packages are supported.

# Version

- [ ] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links


JLockerman (approved):
So it's install from source if you're not on cloud, then run `CREATE EXTENSION timescaledb_tookit;`? Looks right to me.

PR #176, opened 2021-07-13
New overview page

CoolAssPuppy:
# Description

The overview page was outdated. Cleaned it up to match the www homepage and link to the YouTube channel.

# Version

- [ ] Latest (Default)
- [ ] Version 1.7
- [X] Older [applicable to all versions of TimescaleDB]

# Links


ryanbooz:
So, yeah - we really need to get rid of `master` at this point because it currently stands - it serves no purpose in the release process and isn't being kept up-to-date. Instead, we decided that for now `latest` would effectively be our "master" since it's always the head of documentation. As needed, other version branches are created and we can port backwards for older changes.

This is a workflow/subject that we need to get straight again and make sure it's clear in the contribution docs.

Probably easiest to get a diff of your changes (or just copy what you did), create a new branch off of `latest` and then re-apply and create a new PR. Rebasing this one on `latest` might do it too.

PR #227, opened 2021-08-02
Remove reference to postgis-docker

gayyappan:
We do not build docker images with postgis starting with
timescaledb 2.4.

Version : TimescaleDB 2.4.0 onwards


svenklemm (approved):

Loquacity (approved):
I suspect there are going to be incoming links to this section that might break when this merges. Could be worth pre-emptively searching for them.

jacobprall (approved):

PR #232, opened 2021-08-03
KBase migration - maintenance windows

Loquacity:
# Description

Migrate https://kb.timescale.cloud/en/articles/2752348-details-on-maintenance-operations to docs.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links


timescale-cloud/maintenance.md
```diff
@@ -0,0 +1,51 @@
+# Maintenance
+On Managed Service for TimescaleDB (MST), software updates are handled automatically by us, and you do not need to perform any actions.
```
ryanbooz, line 2:
```suggestion
On Timescale Cloud, software updates are handled automatically by us, and you do not need to perform any actions.
```

timescale-cloud/maintenance.md
```diff
@@ -0,0 +1,51 @@
...
+
+When you are considering your maintenance window schedule, you might prefer to
+choose a day and time that usually has very low activity, such as during the
+early hours of the morning, or over the weekend. This can help minimize the
+impact of a short service interruption. Alternatively, you might prefer to have
+your maintenance window occur during office hours, so that you can monitor your
+system during the upgrade.
+
+### Procedure: Adjusting your maintenance window
+1.  [Log in to your MST account][mst-login]. Click the name of the service that
```
ryanbooz, line 29:
```suggestion
1.  [Log in to your Timescale Cloud account][mst-login]. Click the name of the service that
```

ryanbooz (changes requested):

timescale-cloud/maintenance.md
```diff
@@ -0,0 +1,51 @@
+# Maintenance
+On Timescale Cloud, software updates are handled automatically by us, and you do not need to perform any actions.
+
+Non-critical software updates are applied during a maintenance window that you can define to suit your workload. If we detect a security vulnerability that affects you, we might need to perform maintenance outside of the scheduled maintenance window.
+
+<highlight type="important">
+During a maintenance update, the DNS name remains the same, but the IP address
```
camscale, line 7: I would change "During" to "After". "During" suggests that the IP address is different only during the maintenance window, but I assume that the restart that happens as a result of maintenance means it now has a different IP address going forward (until the next maintenance).

timescale-cloud/maintenance.md
```diff
@@ -0,0 +1,51 @@
...
+    you want to manage the maintenance window for.
+1.  In the `Maintenance window` section, click `Change`.
+1.  In the `Service Maintenance Window` dialog, select the day of the week and
+    the time (in Universal Coordinated Time) you want the maintenance window to
+    start. Maintenance windows can run for up to four hours.
+    <img class="main-content__illustration" src="https://s3.amazonaws.com/assets.timescale.com/docs/images/mst-maintwindow.png" alt="Adjust maintenance window"/>
+1.  Click `Save Changes`.
+
+## Critical updates
+Critical upgrades and security fixes can be installed outside normal maintenance windows when necessary, and sometimes require have a short outage.
```
camscale, line 39: remove "have" - "require have a short outage" -> "require a short outage".

camscale, line 39: change "can" to "may": "security fixes can be installed outside normal maintenance windows" -> "security fixes may be installed outside normal maintenance windows". The former implies it is an action that they may take, the latter implies it is something that happens regardless.

camscale (commented):
Just a couple of language suggestions. I'm no expert here so feel free to ignore.

timescale-cloud/maintenance.md
```diff
@@ -0,0 +1,51 @@
+# Maintenance
+On Timescale Cloud, software updates are handled automatically by us, and you do not need to perform any actions.
+
+Non-critical software updates are applied during a maintenance window that you can define to suit your workload. If we detect a security vulnerability that affects you, we might need to perform maintenance outside of the scheduled maintenance window.
+
+<highlight type="important">
+During a maintenance update, the DNS name remains the same, but the IP address
+it points to changes.
```
Loquacity, line 8:
```suggestion
After a maintenance update, the DNS name remains the same, but the IP address
it points to changes.
```

timescale-cloud/maintenance.md
Loquacity:
```suggestion
Critical upgrades and security fixes can be installed outside normal maintenance windows when necessary, and sometimes require a short outage.
```

Loquacity: I'm not sure how you're drawing that difference between "can" and "may" honestly 😂  Generally speaking, they mean exactly the same thing, but "may" is more formal. In my own writing practice, I tend to avoid "may" because it's a bit more ambiguous than some of the alternatives (it can mean that you have permission to do something conferred by someone else, or that you have the ability in yourself to do something), and it just reads a bit too stiff and formal for the writing style I am using here. In this case, I think if you want to make the sentence more assertive, using "are" would work better?

camscale: Fair enough. I agree what you say, but for some reason when I read "can be installed", I read it as giving the user the option, rather than stating it may be done. With "may be installed" I read it as something that could happen. But I also do get that they are really the same.

:+1: to using "are" instead of "can be"

timescale-cloud/maintenance.md
```diff
@@ -0,0 +1,51 @@
...
+    you want to manage the maintenance window for.
+1.  In the `Maintenance window` section, click `Change`.
+1.  In the `Service Maintenance Window` dialog, select the day of the week and
+    the time (in Universal Coordinated Time) you want the maintenance window to
+    start. Maintenance windows can run for up to four hours.
+    <img class="main-content__illustration" src="https://s3.amazonaws.com/assets.timescale.com/docs/images/mst-maintwindow.png" alt="Adjust maintenance window"/>
+1.  Click `Save Changes`.
+
+## Critical updates
+Critical upgrades and security fixes can be installed outside normal maintenance windows when necessary, and sometimes require a short outage.
```
Loquacity, line 39:
```suggestion
Critical upgrades and security fixes are installed outside normal maintenance windows when necessary, and sometimes require a short outage.
```

timescale-cloud/maintenance.md
```diff
@@ -0,0 +1,52 @@
+# Maintenance
+On Timescale Cloud, software updates are handled automatically by us, and you do not need to perform any actions.
+
+Non-critical software updates are applied during a maintenance window that you can define to suit your workload. If we detect a security vulnerability that affects you, we might need to perform maintenance outside of the scheduled maintenance window.
```
Loquacity, line 4:
```suggestion
Non-critical software updates are applied during a maintenance window that you can define to suit your workload. 
If we detect a security vulnerability that affects you, we might need to perform maintenance outside of 
the scheduled maintenance window.
```

ryanbooz (approved):

PR #272, opened 2021-08-19
Move experimental features

Loquacity:
# Description

Move experimental features out of the 'experimental' directory, to their correct locations. Add the `experimental` tag (https://github.com/timescale/web-documentation/pull/220) and admonitions.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

### DIFF
--- a/api/page-index/page-index.js
+++ b/api/page-index/page-index.js
@@ -126,6 +126,14 @@ module.exports = [
           {
             title: "set_replication_factor",
             href: "set_replication_factor"
+          },
+          {
+            title: "copy_chunk",
+            href: "copy_chunk_experimental"
+          },
+          {
+            title: "move_chunk",
+            href: "move_chunk_experimental"
           }
         ]
       },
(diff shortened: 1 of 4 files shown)


api/page-index/page-index.js
```diff
@@ -274,6 +282,10 @@ module.exports = [
               {
                 title: "interpolate",
                 href: "interpolate"
+              },
```
mfreed, line 285: Should we put `time_bucket_ng` right under `time_bucket` above?  (That feels right to me?)

If you are instead trying to put all experimental features at the end of the section -> then just one issue is I think this is incorrectly placed as a child of "gapfilling" in the hierarchy -> needs to be moved up one level.

mfreed (approved):

api/page-index/page-index.js
Loquacity: Ah, good spot!

Loquacity:
@jacobprall I'm not seeing the experimental tag? Is this because the test environment isn't using the latest code, or have I done something wrong?

image

jacobprall (commented):
I've updated the front end to handle multiple tags (Community + Experimental), should be good to go!

PR #314, opened 2021-08-30
Small line edits to time_bucket_ng API

Loquacity:
# Description

Just some general tidy up.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Related to: https://github.com/timescale/docs/pull/308

### DIFF
--- a/api/time_bucket_ng.md
+++ b/api/time_bucket_ng.md
@@ -1,9 +1,9 @@
 ## timescaledb_experimental.time_bucket_ng() <tag type="experimental">Experimental</tag>
-
-The `time_bucket_ng()` (next generation) experimental function is
-similar to [`time_bucket()`][time_bucket], but works with years and
-months. It is expected that the feature will also support timezones
-in a future release.
+The `time_bucket_ng()` (next generation) experimental function is an updated
+version of  the original [`time_bucket()`][time_bucket] function. While
+`time_bucket` works with small units of time, the `time_bucket_ng()` function
+uses years and months. The `time_bucket_ng()` function does not, at this stage,
+support timezones.
 
 <highlight type="warning">
 Experimental features could have bugs! They might not be backwards compatible,
@@ -32,8 +32,8 @@ SELECT timescaledb_experimental.time_bucket_ng('1 year', date '2021-08-01');
 (1 row)
 ```
 
-To split time into buckets we use a starting point in time called `origin`. The
-default origin is '2000-01-01'. `time_bucket_ng` can't work with timestamps
+To split time into buckets, use a starting point in time called `origin`. The
+default origin is `2000-01-01`. `time_bucket_ng` cannot use timestamps
 earlier than `origin`:
 
 ```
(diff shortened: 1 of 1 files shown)


api/time_bucket_ng.md
```diff
@@ -1,9 +1,9 @@
 ## timescaledb_experimental.time_bucket_ng() <tag type="experimental">Experimental</tag>
-
-The `time_bucket_ng()` (next generation) experimental function is
-similar to [`time_bucket()`][time_bucket], but works with years and
-months. It is expected that the feature will also support timezones
-in a future release.
+The `time_bucket_ng()` (next generation) experimental function is an updated
+version of  the original [`time_bucket()`][time_bucket] function. While
+`time_bucket` works with small units of time, the `time_bucket_ng()` function
```
gayyappan, line 4: I do not think this is accurate. @afiskon Could you confirm if time_bucket_ng supports all the "intervals" supported by time_bucket + ( years  and months ) ?
@Loquacity See the comparison table between time_bucket and time_bucket_ng here: https://blog.timescale.com/blog/move-fast-but-dont-break-things-introducing-the-experimental-schema-with-new-experimental-features-in-timescaledb-2-4/

api/time_bucket_ng.md
```diff
@@ -32,24 +32,21 @@ SELECT timescaledb_experimental.time_bucket_ng('1 year', date '2021-08-01');
 (1 row)
 ```
 
-To split time into buckets we use a starting point in time called `origin`. The
-default origin is '2000-01-01'. `time_bucket_ng` can't work with timestamps
-earlier than `origin`:
+To split time into buckets, `time_bucket_ng()` uses a starting point in time
+called `origin`. The default origin is `2000-01-01`. `time_bucket_ng` cannot use
+timestamps earlier than `origin`:
```
gayyappan, line 37: I think some explanation is required here. I am not sure why the example below does not work . @afiskon  Could you add some text here that describes why the example fails?

api/time_bucket_ng.md
afiskon: time_bucket_ng() does support everything time_bucket() does. There could be a mistake in the blogpost though. Not sure what was meant by the support of timezones in time_bucket(). Last time I reviewed the draft the table looked different.

api/time_bucket_ng.md
afiskon: The reason is that the timestamp is earlier than the default origin, as stated in the sentence. Any ideas on how to clarify this?

afiskon (approved):
Looks OK, thanks!

api/time_bucket_ng.md
afiskon: cc: @ryanbooz

api/time_bucket_ng.md
gayyappan: I think part of the confusion is that we do not document what the argument names are. I wasn't sure what date meant here.  See: https://docs.timescale.com/api/latest/hyperfunctions/time_bucket/#time-bucket. 

This api section defintely needs a required + optional arguments section.
@afiskon We should also revise the error message in the source code and use the correct argument names.

afiskon:
@Loquacity just wanted to let you know: during the last team meeting, @gayyappan and I agreed that this PR can be merge in the current state. This feature is in active development and going to change, and the documentation is going to change accordingly. I will address the issues @gayyappan pointed out in the follow-up pull requests.

PR #327, opened 2021-09-01
Add "note" as possible highlight type in readme

zseta:
Also not sure if there are more highlight types that could be used besides these?


Loquacity:
I was hoping for "note" (green), "important" (amber), and "warning" (red), actually.

jacobprall (commented):
Both note and tip are acceptable and show up in blue, important in gold and warning in red

Loquacity:
> Both note and tip are acceptable and show up in blue, important in gold and warning in red

In that case, this edit to the readme should say that.

PR #484, opened 2021-09-28
Toolkit - counter agg content

Loquacity:
# Description

Adds counter agg Toolkit content

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Partial: https://github.com/timescale/docs-private/issues/4

### DIFF
--- a/timescaledb/how-to-guides/toolkit/counter_agg.md
+++ b/timescaledb/how-to-guides/toolkit/counter_agg.md
@@ -1,17 +1,32 @@
 # Counter aggregates
-There are two types of metrics that can be recorded in a database: gauges, and counters. Gauges fluctuate up and down, like temperature or speed, while counters always increase, like the total number of visitors to a website.
+There are two types of metrics that can be recorded in a database: gauges, and
+counters. Gauges fluctuate up and down, like temperature or speed, while
+counters always increase, like the total number of visitors to a website.
 
-When you process counter data, it is usually assumed that if the value of the counter goes down, the counter has been reset. For example, if you wanted to count how many times a runner ran around a track, you would expect the values to continuously increase: 1, 2, 3, 4, and so on. If the counter reset to 0, you would expect that this was a new runner starting on their first lap, not the original runner going back to 1. This can become a problem if you want to continue counting from where you left off, rather than resetting to 0. A reset could occur if you have had a short server outage, or any number of other reasons. To get around this, you can analyze counter data by taking its change over a time period, which accounts for resets.
+When you process counter data, it is usually assumed that if the value of the
+counter goes down, the counter has been reset. For example, if you wanted to
+count how many times a runner ran around a track, you would expect the values to
+continuously increase: 1, 2, 3, 4, and so on. If the counter reset to 0, you
+would expect that this was a new runner starting on their first lap, not the
+original runner going back to 1. This can become a problem if you want to
+continue counting from where you left off, rather than resetting to 0. A reset
+could occur if you have had a short server outage, or any number of other
+reasons. To get around this, you can analyze counter data by looking at the
+change over time, which accounts for resets.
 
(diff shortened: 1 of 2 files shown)


timescaledb/how-to-guides/toolkit/counter_agg.md
```diff
@@ -0,0 +1,164 @@
+# Counter aggregates
+There are two types of metrics that can be recorded in a database: gauges, and
+counters. Gauges fluctuate up and down, like temperature or speed, while
```
davidkohn88, line 3: There are far more types than that and I wouldn't want to imply that these are the only two types of metrics one can store in a database I think we need to figure out how to limit this to the sort of constrained "metrics in the sense that Prometheus and other APM type things talk about". This feels misleading.

timescaledb/how-to-guides/toolkit/counter_agg.md
```diff
@@ -0,0 +1,164 @@
...
+counters always increase, like the total number of visitors to a website.
+
```
davidkohn88, line 5: Maybe we can take the speed analogy and use this throughout? use a car as our story? To explain them, let's think about a car: we have gauges which can fluctuate up and down, like speed or the fuel gauge, and we have counters like the odometer which always increases and tracks the total number of miles the car has traveled.

timescaledb/how-to-guides/toolkit/counter_agg.md
```diff
@@ -0,0 +1,164 @@
...
+Counter aggregates can be used in continuous aggregates, even though they are
+not parallelizable in PostgreSQL. For more information, see the section on
+[parallelism and ordering][#parallelism-and-ordering].
+</highlight>
+
+## Run a counter aggregate query
+In this procedure, we are using an example table called `example` that contains
+counter data.
+
+### Procedure: Running a counter aggregate query
```
davidkohn88, line 31: I still, really really hate calling these procedures, can we call it an example? The overloading with SQL procedures is really, really confusing to me as a SQL user...

timescaledb/how-to-guides/toolkit/counter_agg.md
```diff
@@ -0,0 +1,164 @@
...
+
+When you process counter data, it is usually assumed that if the value of the
+counter goes down, the counter has been reset. For example, if you wanted to
+count how many times a runner ran around a track, you would expect the values to
+continuously increase: 1, 2, 3, 4, and so on. If the counter reset to 0, you
+would expect that this was a new runner starting on their first lap, not the
+original runner going back to 1. This can become a problem if you want to
+continue counting from where you left off, rather than resetting to 0. A reset
+could occur if you have had a short server outage, or any number of other
+reasons. To get around this, you can analyze counter data by looking at the
```
davidkohn88, line 14: I think introducing the second runner is confusing here? Perhaps we could continue the car analogy? Where if someone were able to reset the odometer (maybe using the "TRIP" button, you can still calculate the overall miles traveled by assuming that each time it resets or decreases, you just keep a running total and treat each time it decreases as a reset. I'm not expressing this well, but maybe it'll spark something for you...

davidkohn88 (commented):
For all of the other ones we've moved this into the function family area of the API guide for now, I think we should do that here, similar to how we've done it for the other functions and then each function have their own place rather than trying to get this whole piece merged, I also think it provides a reasonable separation of concerns. Probably cut down a bit and with the example removed for now, we can then go back and add example usages in a different section.

Loquacity:
> For all of the other ones we've moved this into the function family area of the API guide for now, I think we should do that here, similar to how we've done it for the other functions and then each function have their own place rather than trying to get this whole piece merged, I also think it provides a reasonable separation of concerns. Probably cut down a bit and with the example removed for now, we can then go back and add example usages in a different section.

No, this content doesn't belong in an API doc. It was only put there in the first place as a stop-gap measure while we got the How To in to shape.

Loquacity:
New PR incoming.

PR #562, opened 2021-10-28
remove comment

jacobprall:
# Description

# Version

- [ ] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

### DIFF
--- a/timescaledb/overview/release-notes/index.md
+++ b/timescaledb/overview/release-notes/index.md
@@ -34,9 +34,8 @@ follow these [setup instructions][distributed-hypertables-setup].
 
 You can read more about this release on our [blog post](https://tsdb.co/timescaledb-2-5).
 This release also contains bug fixes since the 2.4.2 release. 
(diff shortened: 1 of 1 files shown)


Loquacity:
> @Loquacity haven't we agreed to keep the 3 status (2 commented out) at all times? @jacobprall what is the purpose of this change?

Yep, that was my understanding. I'm wondering if maybe it broke something somehow? @jacobprall ?

jacobprall:
The mdx parser didn't know how to handle it and thus broke - I'll look into if there's a work around

Loquacity:
> The mdx parser didn't know how to handle it and thus broke - I'll look into if there's a work around

I've found that you need to have blank lines around comments for them to work. Did you try that?

NunoFilipeSantos:
@jacobprall @Loquacity if that's the case maybe leave that as a comment on the file so when we do this in the future we don't forget.

PR #586, opened 2021-11-03
Feature/vale linter

jacobprall:


Loquacity:
Looks like it's working nicely! 🎉

Loquacity (approved):

.github/styles/vocab.txt
```diff
@@ -0,0 +1,12 @@
+eg
```
Loquacity, line 1: Why did you include eg?

jacobprall: It was erroring out on the use of eg for exempli gratia, but we can remove it if we want to target and remove eg

Loquacity: Yeah, I want "for example" everywhere.

Issue #596, opened 2021-11-12
[Site Bug] Changes for the AWS Lambda tutorial [WAS: Issue with the page: /docs-tutorial-aws-lambda/timescaledb/tutorial-aws-lambda/how-to-guides/distributed-hypertables/enable-native-replication/]


tatzlwurm2:
_Use this template for reporting bugs in the docs._

# Describe the bug

Documentation is very confusing:  warning
Native replication is currently under development and lacks functionality for a complete high-availability solution. For instance, if a node fails one might need to (re-)replicate the chunks that had a replica on the failed node. While it is possible to copy chunks from one node to another, this functionality is experimental and not yet automated. For production environments, we therefore recommend keeping the replication factor set at the default value of 1, and instead use streaming replication on each data node.

``` 
 I am doing something wrong?  2 access nodes, 4 data nodes (replication factor 2).  PG 13.4 timescale 2.5.0  I have tried the following
psql cdsdb -U cds_u -c"truncate pgbench_history"
psql cdsdb -U cds_u -c"create table pgbench_history1 ( like pgbench_history including all)"
psql cdsdb -U cds_u -c"drop table pgbench_history "
psql cdsdb -U cds_u -c"alter table pgbench_history1 rename to pgbench_history"
psql cdsdb -U cds_u -c"SELECT create_distributed_hypertable('pgbench_history','mtime',chunk_time_interval => INTERVAL '1 day')"
psql cdsdb -U cds_u -c"SELECT set_replication_factor('pgbench_history', 2)"
psql cdsdb -c"select pg_stat_statements_reset()"
pgbench -c20 -j4 -T3600 -rn -P60 -p5600 cdsdb -U cds_u
psql cdsdb -c"create table pg_stat_hypertable_rep2-failtest1 as select * from pg_stat_statements;"
psql cdsdb -c"select pg_stat_statements_reset()"
psql cdsdb -U cds_u -c"select * from timescaledb_information.chunks ;"
During this test I see
cdsdb=> select * from timescaledb_information.chunks ;
 hypertable_schema | hypertable_name |     chunk_schema      |       chunk_name       | primary_dimension |   primary_dimension_type    |      range_start       |       range_end        | range_start_integer | range_end_integer | is_compressed | chunk_
tablespace |       data_nodes
-------------------+-----------------+-----------------------+------------------------+-------------------+-----------------------------+------------------------+------------------------+---------------------+-------------------+---------------+-------
-----------+-------------------------
[...]

PR #601, opened 2021-11-16
Improve the multi-node HA section

erimatnor:
Make it clear in the multi-node HA section that there are two viable
ways to provide multi-node HA. In particular, we should emphasize that
native replication is still not recommended for production use and
suggest setting up standby nodes as an alternative.

The change also fixes a couple of issues like text that ends in the
middle of the sentence, missing content that existed in old docs, etc.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

Fixes https://github.com/timescale/docs/issues/596

### DIFF
--- a/timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
+++ b/timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
@@ -1,55 +1,131 @@
-# High availability (HA) with multi-node
-A multi-node installation can be configured so that each chunk is written to
-multiple data nodes. This natively replicates data at the chunk level, and
-protects against data node failures. It also provides high availability using an
-alternative to fully replicating each data node using.
+# High availability with multi-node
+A multi-node installation of TimescaleDB can be made highly available
+by setting up one or more standbys for each node in the cluster, or by 
+natively replicating data at the chunk level.
 
-The data nodes themselves require no additional setup to use native replication,
-but the access node relies on streaming replication by default, so it requires
-some configuration to make it highly available.
+Using standby nodes relies on streaming replication and you set it up
+in a similar way to [configuring single-node HA][single-ha], although the
+configuration needs to be applied to each node independently.
 
-When data is written to a chunk, it uses the `replication_factor` setting to
-determine how many nodes it replicates the data to. By default, this is set to
-`1`, so there is no native replication. You can increase this number when you
-create the hypertable. For example, to replicate the data across a total of
-three data nodes:
+To replicate data at the chunk level, you can use the built-in 
+capabilities of multi-node TimescaleDB to avoid having to
+replicate entire data nodes. The access node still relies on a
+streaming replication standby, but the data nodes need no additional
+configuration. Instead, the existing pool of data nodes share
+responsibility to host chunk replicas and handle node failures.
+
+There are advantages and disadvantages to each approach. 
+Setting up standbys for each node in the cluster ensures that 
+standbys are identical at the instance level, and this is a tried
+and tested method to provide high availability. However, it also
+requires more setting up and maintenance for the mirror cluster.
+
+Native replication typically requires less resources, nodes, and
+configuration, and takes advantage of built-in capabilities, such as
+adding and removing data nodes, and different replication factors on
+each distributed hypertable. However, only chunks are replicated on
+the data nodes.
+
+The rest of this section discusses native replication. To set up
+standbys for each node, follow the instructions for [single node
+HA][single-ha].
+
+## Native replication
+Native replication is a set of capabilities and APIs that allow you to
+build a highly available multi-node TimescaleDB installation. At the
+core of native replication is the ability to write copies of a chunk
+to multiple data nodes in order to have alternative _chunk replicas_
+in case of a data node failure. If one data node fails, its chunks
+should be available on at least one other data node. If a data node is
+permanently lost, a new data node can be added to the cluster, and
+lost chunk replicas can be re-replicated from other data nodes to
+reach the number of desired chunk replicas.
+
+<highlight type="warning"> 
+Native replication in TimescaleDB is under development and 
+currently lacks functionality for a complete high-availability
+solution. Some functionality described in this section is still
+experimental. For production environments, we recommend setting up
+standbys for each node in a multi-node cluster.
+</highlight>
+
+### Automation
+Similar to how high-availability configurations for single-node
+PostgreSQL uses a system like Patroni for automatically handling
+fail-over, native replication requires an external entity to
+orchestrate fail-over, chunk re-replication, and data node
+management. This orchestration is _not_ provided by default in
+TimescaleDB and therefore needs to be implemented separately. The
+sections below describe how to enable native replication and the steps
+involved to implement high availability in case of node failures.
+
+### Configuring native replication
+The first step to enable native replication is to configure a standby
+for the access node. This process is identical to setting up a [single
+node standby][single-ha].
+
+The next step is to enable native replication on a distributed
+hypertable. Native replication is governed by the
+`replication_factor`, which determines how many data nodes a chunk is
+replicated to. This setting is configured separately for each
+hypertable, which means the same database can have some distributed
+hypertables that are replicated and others that are not.
+
+By default, the replication factor is set to `1`, so there is no
+native replication. You can increase this number when you create the
+hypertable. For example, to replicate the data across a total of three
+data nodes:
 ```sql
 SELECT create_distributed_hypertable('conditions', 'time', 'location',
 	replication_factor => 3);
 ```
 
-Alternatively, you can set the `replication_factor` setting on an existing
-hypertable in a multi-node environment using the
-[`set_replication_factor`][set_replication_factor] call.
+Alternatively, you can use the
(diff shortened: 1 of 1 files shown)


timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
...
+similar to [configuring single-node HA][single-ha], although the
+configuration needs to be applied to each node independently.
+
+The second option relies on built-in capabilities of multi-node
+TimescaleDB to replicate data at the chunk level instead of
+replicating entire data nodes. The access node still relies on a
+streaming replication standby, but the data nodes need no additional
+configuration. Instead, the existing pool of data nodes share
+responsibility to host chunk replicas in order to be able to sustain
+node failures.
```
Loquacity, line 19:
```suggestion
To replicate data at the chunk level, you can use the built-in 
capabilities of multi-node TimescaleDB to avoid having to
replicate entire data nodes. The access node still relies on a
streaming replication standby, but the data nodes need no additional
configuration. Instead, the existing pool of data nodes share
responsibility to host chunk replicas and handle node failures.
```

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
...
+
+There are pros and cons of each approach. Setting up standbys for each
+node in the cluster ensures that standbys are identical at the
+instance level and is a tried and tested method to provide high
+availability. However, it is also a heavy-weight solution and
+essentially requires setting up a mirror cluster.
```
Loquacity, line 25:
```suggestion
There are advantages and disadvantages to each approach. 
Setting up standbys for each node in the cluster ensures that 
standbys are identical at the instance level, and this is a tried
and tested method to provide high availability. However, it also
requires more setting up and maintenance for the mirror cluster.
```

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
...
+
+Native replication, on the other hand, offers simpler configuration
+and additional, built-in, capabilities, like being able to easier add
+and remove data nodes, and use different replication factors on each
+distributed hypertable. The downside, however, is that only chunks are
+replicated on the data nodes.
```
Loquacity, line 31:
```suggestion
Native replication requires less configuration, and takes advantage of 
built-in capabilities, such as adding and removing data nodes, and 
different replication factors on each distributed hypertable. However, 
only chunks are replicated on the data nodes.
```

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
...
+
+The rest of this section focuses on native replication. To setup
+standbys for each node, follow the instructions for [single node
+HA][single-ha].
```
Loquacity, line 35:
```suggestion
The rest of this section discusses native replication. To set up
standbys for each node, follow the instructions for [single node
HA][single-ha].
```

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
 # High availability (HA) with multi-node
```
Loquacity, line 1:
```suggestion
# High availability with multi-node
```

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
...
+
+Native replication is a set of capabilities and APIs around which one
+can build a highly available multi-node TimescaleDB installation. At
+the core of native replication is the ability to write copies
+(replicas) of a chunk to multiple data nodes. Thus, a failure of one
+data node should not lead to loss of data as copies exist on other
+data nodes. If a data node is permanently lost, a new data node can be
+added to the cluster, and lost chunk replicas can be copied
+(re-replicated) from other data nodes to again reach the number of
+desired chunk replicas.
```
Loquacity, line 47:
```suggestion
## Native replication
Native replication is a set of capabilities and APIs that allow you to 
build a highly available multi-node TimescaleDB installation. At
the core of native replication is the ability to write copies or replicas
of a chunk to multiple data nodes. If one data node fails, copies exist
on other data nodes, so that data is not lost. If a data node is 
permanently lost, a new data node can be added to the cluster, 
and lost chunk replicas can be copied or re-replicated from 
other data nodes to reach the number of desired chunk replicas.
```

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
...
+
+Note that orchestration around native replication is not provided as
+part of TimescaleDB itself, so, like standard single-node standby
+setups, external entities need to manage the set of data nodes,
+trigger failover, and re-replicate chunks.
```
Loquacity, line 52:
```suggestion
Native replication is not provided in TimescaleDB by default. You can
set this up in the same way as a standard single-node standby, and
use external entities to manage the set of data nodes, trigger failover, 
and re-replicate chunks. This section contains information about how
to do that.
```

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
...
+
+<highlight type="warning"> Native replication is currently under
+development and lacks functionality for a complete high-availability
+solution. Some functionality described in this section is still
+experimental. For production environments, we recommend setting up
+standbys for each node in a multi-node cluster.</highlight>
```
Loquacity, line 58:
```suggestion
<highlight type="warning"> 
Native replication in TimescaleDB is under development and 
currently lacks functionality for a complete high-availability
solution. Some functionality described in this section is still
experimental. For production environments, we recommend setting up
standbys for each node in a multi-node cluster.
</highlight>
```

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
...
+hypertable. For example, to replicate the data across a total of three
+data nodes:
 ```sql
 SELECT create_distributed_hypertable('conditions', 'time', 'location',
 	replication_factor => 3);
 ```
 
-Alternatively, you can set the `replication_factor` setting on an existing
+Alternatively, you can set the `replication_factor` on an existing
 hypertable in a multi-node environment using the
```
Loquacity, line 83: Turn this into a `<procedure>`

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -1,34 +1,100 @@
...
+handling node failures).
+
+When native replication is enabled, the replication happens whenever
+you ingest data into the table. On each `INSERT` and `COPY` call, each
+row of the data is written to multiple data nodes. This means that you
+don't need to do any extra steps to have newly ingested data
+replicated. When you query replicated data, the query planner only
+includes one replica of each chunk in the query plan.
+
+### Handling node failures
```
Loquacity, line 97:
```suggestion
### Node failures
```

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
```diff
@@ -37,8 +103,11 @@ SELECT delete_data_node('data_node_2', force => true);
 WARNING:  distributed hypertable "conditions" is under-replicated
 ```
 
-If the data node cannot be recovered, native replication allows the node to be
-deleted from the multi-node cluster without losing data.
+If the data node cannot be recovered, native replication allows the
+node to be deleted from the multi-node cluster without losing
+data. Note that you must use the `force` option when deleting the data
+node if the deletion means that the cluster would no longer achieve
+the desired replication factor.
```
Loquacity, line 110:
```suggestion
If the data node cannot be recovered, native replication allows the
node to be deleted from the multi-node cluster without losing
data. Use the `force` option when you delete the data node if 
the deletion means that the cluster no longer achieves the 
desired replication factor.
```

Loquacity (changes requested):

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
erimatnor: "it also requires more setting up" --- sounds a bit weird to me, but I am not a native English speaker. Just want to double-check @Loquacity

timescaledb/how-to-guides/multinode-timescaledb/multinode-ha.md
erimatnor: I think the suggested changes actually subtly changed the meaning of what I was trying to say and is no longer entirely correct. I'll try to tweak my original text to match your suggestion while trying to preserve the original messaging.

Loquacity (approved):

NunoFilipeSantos (approved):
Ship it. :)

PR #638, opened 2021-12-03
Adjust AMI ids for TimescaleDB 2.5.1

svenklemm:
Updated AMI ids for timescaledb 2.5.1 release.


NunoFilipeSantos (approved):
that is actually better!

PR #655, opened 2021-12-14
Document real-time aggregates limitation

afiskon:
# Description

There is a well-known real-time aggregates limitation that is not very well documented and thus can be surprising to some users. This PR documents this limitation.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Fixes https://github.com/timescale/timescaledb/issues/3877


timescaledb/how-to-guides/continuous-aggregates/real-time-aggregates.md
```diff
@@ -28,4 +28,92 @@ You can enable and disable real time aggregation by setting the
...
+SELECT * FROM conditions_summary ORDER BY bucket;
+  city  |   bucket   | min | max
+--------+------------+-----+-----
+ Moscow | 2021-06-14 |  22 |  35
+ Moscow | 2021-06-21 |  31 |  34
+```
+
+In other words, `refresh_continuous_aggregate()` is supposed to be called for the
+data that is not going to be updated anymore. When it's necessary to update the
+data for the past, it's required to explicitly call `refresh_continuous_aggregate()`
```
mfundul, line 116: It's not only data in the past. It may very well be data in the present or the future that end up falling within the current bucket.

afiskon: 👍 changed the wording

timescaledb/how-to-guides/continuous-aggregates/real-time-aggregates.md
```diff
@@ -28,4 +28,92 @@ You can enable and disable real time aggregation by setting the
...
+  ('2021-06-18', 'Moscow', 27),
+  ('2021-06-19', 'Moscow', 28),
+  ('2021-06-20', 'Moscow', 30),
+  ('2021-06-21', 'Moscow', 31),
+  ('2021-06-22', 'Moscow', 34),
+  ('2021-06-23', 'Moscow', 34),
+  ('2021-06-24', 'Moscow', 34),
+  ('2021-06-25', 'Moscow', 32),
+  ('2021-06-26', 'Moscow', 32),
+  ('2021-06-27', 'Moscow', 31);
```
mfundul, line 62: This example does not show the full picture. If you remove `('2021-06-27', 'Moscow', 31)` from conditions and insert it later (instead of just updating old data) it will show the users that even fresh data are not visible by real-time aggregates.

afiskon: The freshness is determined by the timestamp, not by the time WHEN the data was inserted. Feel free to suggest a better wording in the text.

mfundul: It's the timestamp I'm talking about. This example doesn't show that even the freshest timestamp would not be visible.

afiskon: Please take a look at the updated text. I hope it's better now. In this example only one bucket is refreshed, thus if you insert now() it will be visible in the CAGG.

timescaledb/how-to-guides/continuous-aggregates/real-time-aggregates.md
```diff
@@ -28,4 +28,94 @@ You can enable and disable real time aggregation by setting the
 
 </procedure>
 
+## Limitations
```
erimatnor, line 31: IMO, this example is overly long and I am not sure the use will be able to understand exactly what the take-away is. 

I also think we need to update the description of real-time aggregates above. The misleading description needs to be fixed. In particular this:

```
Real time aggregates use the aggregated data and add the
most recent raw data to it to provide accurate and up to date results, without
needing to aggregate data as it is being written.
```

afiskon: This description seems accurate 🤷‍♂️  Speaking of the example, I don't see a way to make it shorter, but feel free to suggest an alternative code. The take-aways are in the text, so the user can just skip the example.

gayyappan: I agree with Erik.  This section describes the problem at length but we don't describe any steps on how to mitigate the problem. The emphasis should be on how to get expected behavior (setting up policies appropriately) +  gotchas (manual refresh calls that can interfere with the materialized boundaries) and use the example within this context.

Loquacity (changes requested):
Please do not add a limitations section. Add it to the existing troubleshooting section, and make sure you follow the format defined on that page.

afiskon:
@Loquacity good point. Please see the updated version of the PR. Feel free to modify it as you see fit.

Loquacity (approved):

PR #721, opened 2022-01-14
Fix Promscale command

ramonguiu:
# Description

Add parameters for host and database which are needed for folks that don't run Promscale and. the database on the same host (the majority).

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]


JamesGuthrie (approved):
LGTM

Loquacity (approved):

PR #751, opened 2022-01-28
Fix miscellaneous typos and emojis

charislam:
# Description

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

### DIFF
--- a/timescaledb/contribute-to-docs.md
+++ b/timescaledb/contribute-to-docs.md
@@ -86,16 +86,16 @@ that case, feel free to ignore them.
 *Adverbs*
 : Do not use.
 
-:+1: Install TimescaleDB.
+&#128077; Install TimescaleDB.
 
-:x: Simply install TimescaleDB.
+&#10060; Simply install TimescaleDB.
 
 *And/Or*
 : Do not use. You can usually pick one. If you're not sure, pick "and".
 
-:x: I like apples and/or oranges.
+&#10060; I like apples and/or oranges.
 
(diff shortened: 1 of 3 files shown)


timescaledb/contribute-to-docs.md
```diff
@@ -86,16 +86,16 @@ that case, feel free to ignore them.
 *Adverbs*
 : Do not use.
 
-:+1: Install TimescaleDB.
+&#128077; Install TimescaleDB.
```
Loquacity, line 89: @jacobprall What's best practice here? I thought we discussed using the HTML codes previously and Gatsby broke them, but they seem to be working now?

charislam: Happy to do these any way we agree on but the Github-style ones don't currently display as emojis :) Unless the decision was that we'd prefer to leave them as plaintext?

jacobprall: Interesting. Not seeing any errors in the logs for this deploy, and they appear to be working on staging - no reason to think they wouldn't work in prod. I'd say the html codes are the way to go - but I will keep an eye on deploys to make sure they don't break any deploys

jacobprall (approved):

timescaledb/contribute-to-docs.md
Loquacity: Thanks Jacob.

PR #771, opened 2022-02-03
Add linting rules

charislam:
# Description

- Prevent some overenthusiastic heading-capitalization warnings
- Add rule to remove 'please' from docs
- Check for broken Markdown reference-style links

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

### DIFF
--- a/.github/styles/Google/MarkdownLinks.yml
+++ b/.github/styles/Google/MarkdownLinks.yml
@@ -1,0 +1,7 @@
+extends: conditional
+message: "Remember to include the link."
+level: error
+ignorecase: false
+# Ensures that the existence of 'first' implies the existence of 'second'.
+first: '(?:\[[\w-]+\])+(\[[\w-]+\])'
+second: '(\[[\w-]+\])+(?:: [\w/:?-]+$)'
--- a/.github/styles/Google/MarkdownLinks.yml
+++ b/.github/styles/Google/MarkdownLinks.yml
@@ -3,5 +3,5 @@ message: "Remember to include the link."
 level: error
 ignorecase: false
 # Ensures that the existence of 'first' implies the existence of 'second'.
-first: '(?:\[[\w-]+\])+(\[[\w-]+\])'
-second: '(\[[\w-]+\])+(?:: [\w/:?-]+$)'
+first: '(?:\[[\w\\`",.() -]+\])(\[[\w-]+\])'
+second: '(\[[\w-]+\])(?:: [\w\/.&#=:?-]+$)'
--- a/.github/styles/Google/Headings.yml
+++ b/.github/styles/Google/Headings.yml
@@ -20,8 +20,11 @@ exceptions:
   - macOS
   - Marketplace
   - MongoDB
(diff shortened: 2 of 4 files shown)


Loquacity:
I don't think we should use 'please'. Generally speaking, it weakens the message, and makes our writing less direct.

jacobprall (approved):

.github/styles/Google/MarkdownLinks.yml
```diff
@@ -0,0 +1,7 @@
+extends: conditional
+message: "Remember to include the link."
+level: error
+ignorecase: false
+# Ensures that the existence of 'first' implies the existence of 'second'.
+first: '(?:\[[\w-]+\])+(\[[\w-]+\])'
+second: '(\[[\w-]+\])+(?:: [\w/:?-]+$)'
```
charislam, line 7: @jacobprall I noticed many of our Markdown reference-style links are broken because there's actually no reference, or the reference text doesn't match.

I _tried_ to write a rule to catch those, but regex is not my strong suit, and I found the Vale docs on conditional rules particularly unenlightening. Does this actually do what I want it to do?

Loquacity (approved):
Conditional on if @jacobprall agrees on the regex.

.github/styles/Google/MarkdownLinks.yml
charislam: Looking at it again I think it should say `second: '(\[[\w-]+\])+(?:: [\w/:?=#-]+$)'` (added `=` and `#` to the url bit)

jacobprall: Try running your regex through https://regex101.com/ and testing there - I think you might need to add an escape on one of the delimiters in the second line if I'm not mistaken. (I'm a little rusty on regex) first line looks good tho

charislam: That site is super helpful!

.github/styles/Google/MarkdownLinks.yml
```diff
@@ -0,0 +1,7 @@
+extends: conditional
+message: "Remember to include the link."
+level: error
+ignorecase: false
+# Ensures that the existence of 'first' implies the existence of 'second'.
+first: '(?:\[[\w\\`",.() -]+\])(\[[\w-]+\])'
+second: '(\[[\w-]+\])(?:: [\w\/.&#=:?-]+$)'
```
charislam, line 7: Fixed these and tested (realized I was missing a lot of possible link text and URL symbols). Here are the test results @Loquacity @jacobprall . Does that look good to you or am I missing some cases?

152819646-341b6e36-7469-4808-8855-bdd4e7ed82e2
152821881-fb996a93-5269-4d1d-902c-293e5c6a88dc

jacobprall: LGTM!

Loquacity: 🚢 it!

PR #779, opened 2022-02-07
Multiple databases

ggodeke:
# Description

Added clarification that Timescale Cloud does not support multiple databases within a single service.  

# Version

- [X] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

### DIFF
--- a/cloud/customize-configuration.md
+++ b/cloud/customize-configuration.md
@@ -63,15 +63,18 @@ and the `Apply changes` (or `Restart and apply changes`) button is available to
 click, prompting you to confirm any changes before the Service is modified.
 
 ## Multiple databases
-Today, Timescale Cloud does not allow the creation of multiple databases within 
-a single service. If you require multiple databases, you will need to create a 
-service for each database. Another option would be to redesign your table 
-structure. Timescale Cloud does not support multiple databases within the same 
-service to reduce complexity. Each database would require its own installation 
-of TimescaleDB, managed independently, which breaks our ability to manage 
-maintenance upgrades for you. Supporting multiple databases also adds 
-significant complexity to the UI. To keep our services as easy to use as 
-possible, we have opted to limit services to one database per service.  
+To create more than one Timescale Cloud database, you need to create a new 
+service for each database. Timescale Cloud does not support multiple 
+databases within the same service. Having a separate service for each database 
+affords each database its own isolated resources. 
+
+Another option is to use [schemas](https://www.postgresql.org/docs/current/ddl-schemas.html). Schemas provide a way to organize 
+tables into logical groups. A single database can contain multiple schemas, 
+which in turn contain tables. The main difference between isolating with 
+databases versus schemas is that a user can access objects in any of the 
+schemas in the database they are connected to, so long as they have the 
+corresponding privileges. Schemas can help isolate smaller use cases 
+that do not warrant their own service.
 
 ## PostgreSQL extensions
 You can use PostgreSQL extensions with Timescale Cloud. If you run
--- a/cloud/customize-configuration.md
+++ b/cloud/customize-configuration.md
@@ -67,7 +67,7 @@ Today, Timescale Cloud does not allow the creation of multiple databases within
 a single service. If you require multiple databases, you will need to create a 
 service for each database. Another option would be to redesign your table 
 structure. Timescale Cloud does not support multiple databases within the same 
-instance to reduce complexity. Each database would require its own installation 
+service to reduce complexity. Each database would require its own installation 
 of TimescaleDB, managed independently, which breaks our ability to manage 
 maintenance upgrades for you. Supporting multiple databases also adds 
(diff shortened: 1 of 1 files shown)


mfreed:
Current text suggests using a separate service for each database.  That gets resource isolation (which we should highlight), which "multiple databases per service" doesn't get.

Alternatively, rather than separate their use cases by _database_ , they can isolate by using separate _schema_.  We probably should add that to the text.

cloud/customize-configuration.md
```diff
@@ -62,6 +62,17 @@ As with the basic database configuration parameters, any changes are highlighted
...
+## Multiple databases
+Today, Timescale Cloud does not allow the creation of multiple databases within 
+a single service. If you require multiple databases, you will need to create a 
+service for each database. Another option would be to redesign your table 
+structure. Timescale Cloud does not support multiple databases within the same 
+service to reduce complexity. Each database would require its own installation 
+of TimescaleDB, managed independently, which breaks our ability to manage 
+maintenance upgrades for you. Supporting multiple databases also adds 
+significant complexity to the UI. To keep our services as easy to use as 
+possible, we have opted to limit services to one database per service.
```
Loquacity, line 74:
```suggestion
To create more than one Timescale Cloud database, you need to create a new 
service for each database. Timescale Cloud does not support multiple 
databases within the same service. 
```

Loquacity (changes requested):

cloud/customize-configuration.md
Loquacity: I would really like to have some actual instructions for "redesign your table structure", because that's not really descriptive enough. I'm also not convinced we need all the internal reasons for designing it this way, it reads a little like "we found this difficult, so we didn't do it".

Loquacity: Additionally, if the content is really just "use one service per db" then it should be an admonition, not under its own heading.

ggodeke:
Thank you both for the feedback!

cloud/customize-configuration.md
ggodeke: Incorporated these changes into my next commit! Let me know if you feel this isn't sufficient. I agree that I am not sure that this warrants its own section.

cloud/customize-configuration.md
```diff
@@ -62,6 +62,20 @@ As with the basic database configuration parameters, any changes are highlighted
...
+databases within the same service. Having a separate service for each database 
+affords each database its own isolated resources. 
+
+Another option is to use [schemas](https://www.postgresql.org/docs/current/ddl-schemas.html). Schemas provide a way to organize 
+tables into logical groups. A single database can contain multiple schemas, 
+which in turn contain tables. The main difference between isolating with 
+databases versus schemas is that a user can access objects in any of the 
+schemas in the database they are connected to, so long as they have the 
+corresponding privileges. Schemas can help isolate smaller use cases 
+that do not warrant their own service.
```
Loquacity, line 77:
```suggestion
Another option is to use 
[schemas](https://www.postgresql.org/docs/current/ddl-schemas.html). 
Schemas provide a way to organize tables into logical groups. A single 
database can contain multiple schemas, which in turn contain tables. 
The main difference between isolating with databases versus schemas 
is that a user can access objects in any of the schemas in the database 
they are connected to, so long as they have the corresponding privileges. 
Schemas can help isolate smaller use cases that do not warrant their 
own service.
```

Loquacity: Instead of all this info about what a schema is, why not just link out to our own Schema how to? (I note you've linked to the pg docs, not internally). Then, you don't need to explain what schemas are, necessarily, you can just say something like "if you want to run multiple datasets within a single service, you can use schemas within a single database. For more information about schemas, see the [Schema documentation](https://docs.timescale.com/timescaledb/latest/how-to-guides/schema-management/)" (or similar).

Loquacity (changes requested):

cloud/customize-configuration.md
```diff
@@ -62,6 +62,22 @@ As with the basic database configuration parameters, any changes are highlighted
...
+
+Another option is to use 
+[schemas](https://www.postgresql.org/docs/current/ddl-schemas.html). 
+Schemas provide a way to organize tables into logical groups. A single 
+database can contain multiple schemas, which in turn contain tables. 
+The main difference between isolating with databases versus schemas 
+is that a user can access objects in any of the schemas in the database 
+they are connected to, so long as they have the corresponding privileges. 
+Schemas can help isolate smaller use cases that do not warrant their 
+own service.
```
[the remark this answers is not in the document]
ggodeke: Sorry, I don't follow. Aren't our docs only talking about schemas in the other sense of the word? I don't think we have docs on schemas in this sense.

Apologies if I am missing something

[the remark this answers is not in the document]
Loquacity: In that case, we should probably add it 😉

PR #800, opened 2022-02-10
Clarify that all hypertable partitioning columns must be included in the primary key

### LINKED ISSUE #764: [Docs RFC] Mention that primary key must contain all partitioning columns for hypertable
# Describe change in content, appearance, or functionality

Modify how-to guide and API reference for hypertables to mention that the partitioning column for the hypertable must be included in the primary key.

If not, you get the error `ERROR: cannot create a unique index without the column "<column_name>" (used in partitioning)`

[Original Slack message](https://timescaledb.slack.com/archives/C63MYDZ35/p1643701908515669)

Relevant pages:
https://docs.timescale.com/api/latest/hypertable/create_hypertable/
https://docs.timescale.com/timescaledb/latest/how-to-guides/hypertables/create/

# Subject matter expert (SME)

@jfjoly 

# Deadline

[When does this need to be addressed]

# Any further info

[Anything else you want to add, or further links]

charislam:
# Description

Clarify existing docs on creating a hypertables. Docs don't explicitly state that the partitioning columns must be included in the primary key, which leads to a confusing user error.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Fixes #764


api/create_hypertable.md
```diff
@@ -232,3 +240,13 @@ important to create space partitioning, see
 [create_distributed_hypertable](/distributed-hypertables/create_distributed_hypertable/).
 The rare cases in which space partitions may be useful for non-distributed
 hypertables are described in the [add_dimension](/hypertable/add_dimension/) section.
+
+### Troubleshooting
+
```
Loquacity, line 245:
```suggestion
```

api/create_hypertable.md
```diff
@@ -232,3 +240,13 @@ important to create space partitioning, see
...
+You may get the following error if your primary key doesn't contain all your
+partitioning columns. Add your time column and `partitioning_column` to your
+primary key and try creating the hypertable again.
+
+```
+ERROR: cannot create a unique index without the column "<column_name>" (used in partitioning)
+```
```
Loquacity, line 252:
```suggestion
If your primary key doesn't contain all your partitioning columns, you 
might get an error like this:
```sql
ERROR: cannot create a unique index without the column "<column_name>" (used in partitioning)
```
To fix this error, add the time column and the `partitioning_column` to your
primary key, and try creating the hypertable again.
```

Comment (Loquacity) [reply api/create_hypertable.md:252]: I've broken GH markup, so probably don't commit the suggestion through the webui. Nevertheless, the point here is to get the important info up front (as per 'classical' style).

**Review by Loquacity — CHANGES_REQUESTED** [review]

File: api/create_hypertable.md [hunk 3]
```diff
@@ -232,3 +240,13 @@ important to create space partitioning, see
 [create_distributed_hypertable](/distributed-hypertables/create_distributed_hypertable/).
 The rare cases in which space partitions may be useful for non-distributed
 hypertables are described in the [add_dimension](/hypertable/add_dimension/) section.
+
+### Troubleshooting
+
+You may get the following error if your primary key doesn't contain all your
+partitioning columns. Add your time column and `partitioning_column` to your
+primary key and try creating the hypertable again.
```
charislam, line 248:
```suggestion
If your primary key doesn't contain all your partitioning columns, you 
might get an error like this:
```

Loquacity (approved):

api/create_hypertable.md
```diff
@@ -49,6 +49,14 @@ still work on the resulting hypertable.
  If you use `SELECT * FROM create_hypertable(...)` you get the return value formatted as a table with column headings.
 </highlight>
 
+<highlight type="note"> 
+The primary key for your PostgreSQL table must include
+all partitioning columns for your hypertable. Since all TimescaleDB hypertables
+are partitioned by time, your primary key must include the `time` column. If you
+specify a `partitioning_column`, your primary key must include that column as 
+well.
```
davidkohn88, line 57: One thing to note here and possibly clarify as it is a bit confusing at first read is that this does not work in the other direction, ie all the columns of your primary key do not need to be included in the partitioning columns, we really don't want people to think that or do that, it will lead to massive over-partitioning if we're not careful.

Maybe something like: 
"""
If you plan to have a `PRIMARY KEY` or `UNIQUE` constraint on your hypertable, it must include all columns that the hypertable is partitioned by. 

All TimescaleDB hypertables are partitioned by a time_column, therefore, this column must always be included in `PRIMARY KEY`s or `UNIQUE` constraints on a hypertable. If you specify a `partitioning_column`,  you must include that column in `PRIMARY KEY`s as well.

This does not work in the opposite direction, you may include columns that are not `partitioning_column`s or `time_column`s in a `PRIMARY KEY` or `UNIQUE` constraint. 

for further info see: 

And then we should split this out into its own explanation and have an example like: 

Text from above...
For instance, given a table such as 

```
CREATE TABLE ht_test(
ts TIMESTAMPTZ,
user_id BIGINT,
device_id BIGINT,
val1 FLOAT,
val2 FLOAT, 
PRIMARY KEY (device_id, ts));
```
You'll note that it has a `PRIMARY KEY` on `device_id, ts` - this means that at any timestamp `ts` you can only have one row per `device_id`.

This `PRIMARY KEY` limits the allowed partitioning schemes we can use, for instance, we could partition by `ts` only by using this `create_hypertable` call:

`SELECT * FROM create_hypertable('ht_test', 'ts');`
Here the `PRIMARY KEY` has two columns, `device_id` and `ts`, and we are partitioning by `ts` as that is our `time_column`, so this works, all columns we are partitioning by are in the `PRIMARY KEY`, and it includes another column, `device_id`. 

Alternatively, we could include `device_id` as a partitioning column as well:
`SELECT * FROM create_hypertable('ht_test', 'ts', 'partitioning_columns'=>'device_id');` 

Again, this works, as `device_id` and `ts` are in our `PRIMARY KEY` and are our partitioning columns. 

However, we cannot include `user_id` in our partitioning columns if it is not part of our `PRIMARY KEY`, so: 
`SELECT * FROM create_hypertable('ht_test', 'ts', 'partitioning_columns'=>'user_id');` 
will result in an error such as: 

```
 ERROR: cannot create a unique index without the column "<column_name>" (used in partitioning)
 ```

However, you could alter the PRIMARY KEY to include the `user_id` column and use any of the partitioning schemes above. 
"""

**Review by davidkohn88 — CHANGES_REQUESTED** [review]
I think the wording on this note will cause more confusion than it helps, I've proposed alternate language and a clarifying section, but I'm happy to try to make a PR with that if you'd prefer...

File: api/create_hypertable.md (hunk 4 above)
Comment (davidkohn88) [reply api/create_hypertable.md:57]: I also think it would be useful to have a section on *WHY* this is, which I would happily write up or maybe work on a blog post about first if you'd prefer...

File: api/create_hypertable.md [hunk 5]
```diff
@@ -49,6 +49,14 @@ still work on the resulting hypertable.
  If you use `SELECT * FROM create_hypertable(...)` you get the return value formatted as a table with column headings.
 </highlight>
 
+<highlight type="note"> 
+The primary key for your PostgreSQL table must include
+all partitioning columns for your hypertable. Since all TimescaleDB hypertables
```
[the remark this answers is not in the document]
charislam: Good point!

api/create_hypertable.md
charislam: I think "why" would be great! It might work better as a blog post, which we then take the highlights from for the docs. If you're interested in writing it, I'd love to see it.

timescaledb/how-to-guides/hypertables/create.md
```diff
@@ -33,6 +33,17 @@ explicitly time-based, as long as it can increment. For example, a
...
+<highlight type="note"> 
+If your table has a `PRIMARY KEY` or `UNIQUE` constraint, that key or constraint
+must include all columns that are used to partition your hypertable. All
+TimescaleDB hypertables are partitioned by a time column, so the key or
+constraint must include the time column.
+
+It may additionally include columns that aren't partitioning columns. For more
+information, see [the troubleshooting
+section](https://docs.timescale.com/timescaledb/latest/how-to-guides/hypertables/create/#primary-keys-and-unique-constraints).
+</highlight>
```
ryanbooz, line 45: Not a big fan that we repeat this text multiple times (with minor changes). I'd prefer we add this to the section below and make a small note highlight that points to the larger section below so that we only have to edit one place later.

Maybe something like: "If you are trying to create a hypertable with a `PRIMARY KEY` or `UNIQUE` constraint and receiving an error, please see X for more information"

timescaledb/how-to-guides/hypertables/create.md
```diff
@@ -74,9 +85,79 @@ hypertable fails. For more information about setting up multi-node, see the
...
+partitioning) 
+```
+
+To fix such an error, add the time column and any other partitioning columns to
+your `PRIMARY KEY` or `UNIQUE` constraint. In this example, you should add the
+`user_id` column.
 
+[create-hypertable]: /api/:currentVersion/hypertable/create_hypertable/
 [migrate-data]: /how-to-guides/migrate-data
 [postgres-createtable]: https://www.postgresql.org/docs/9.1/sql-createtable.html
```
ryanbooz, line 162:
```suggestion
[postgres-createtable]: https://www.postgresql.org/docs/current/sql-createtable.html
```

api/create_hypertable.md
ryanbooz: Honestly, I think we could add a paragraph (or two at most) at the beginning of the new Troubleshooting section that simply says, "Hypertables use PostgreSQL partitioning and a constraint on doing partitioning effectively is that any unique constraints must be a part of the partitioning keys. This is to prevent unnecessary work across partitions when dealing with lookups, Therefore..." - or whatever makes sense. If you can give us 3-4 sentences @davidkohn88 of what you think would be helpful as a starting place that would be much appreciated!

timescaledb/how-to-guides/hypertables/create.md
```diff
@@ -74,9 +85,79 @@ hypertable fails. For more information about setting up multi-node, see the
...
+section](https://docs.timescale.com/timescaledb/latest/how-to-guides/hypertables/create/#primary-keys-and-unique-constraints).
+</highlight>
+
 </procedure>
 
+## Troubleshooting
+
+### Primary keys and unique constraints
+
+If your `PRIMARY KEY` or `UNIQUE` constraint doesn't contain all your
```
davidkohn88, line 95: I'd have this match the copy from above, I don't like the "if it doesn't you get an error", they're probably here because they got that error they want to know something more. 

The other thing about all this copy that I'm still struggling with is that it is rather confusing in terms of what's included in what. I was trying to get at that with the example below - would something like - columns used for partitioning must be included in any `UNIQUE` constraint(s) / `PRIMARY KEY`s on the hypertable be any better? I just find all this very confusing when I'm reading it and I know the topic very well, so it must be confusing for users? How are y'all feeling about it? Do you follow it or find it a bit opaque?

davidkohn88, line 95: I wonder if we should jsut have a brief section explaining that a PRIMARY KEY includes a UNIQUE constraint, so we can just refer to everything by that, but in case people don't know...

timescaledb/how-to-guides/hypertables/create.md
```diff
@@ -74,9 +85,79 @@ hypertable fails. For more information about setting up multi-node, see the
...
+partitioning columns, you get an error. Partitioning columns always include the
```
davidkohn88, line 96: I'd avoid using the word include here, as it's used in the index context and could cause confusion. 

I might just make a brief separate section ie: 

Which columns are partitioning columns? 
1) The `time_column` used to create the hypertable- every hypertable partitions by its `time_column`. Therefore the `time_column` must be part of any `UNIQUE` constraint / `PRIMARY KEY`. 
2) Any space partitioning columns - space partitions are optional and not included on every hypertable, they may be defined with the `partitioning_columns` parameter of `create_hypertable`, or with the third
argument of `create_distributed_hypertable` - and I'd link to the API docs for those?

timescaledb/how-to-guides/hypertables/create.md
```diff
@@ -74,9 +85,79 @@ hypertable fails. For more information about setting up multi-node, see the
...
+```
+
+You get the following error:
+
+```
+ ERROR: cannot create a unique index without the column "<COLUMN_NAME>" (used in
+partitioning) 
+```
+
+To fix such an error, add the time column and any other partitioning columns to
```
davidkohn88, line 147: This occurs because uniqueness cannot be enforced without having the all partitioning columns in the `PRIMARY KEY` or `UNIQUE` constraint.

To fix...

api/create_hypertable.md
```diff
@@ -17,7 +17,7 @@ still work on the resulting hypertable.
 |Name|Type|Description|
 |---|---|---|
 | `relation` | REGCLASS | Identifier of table to convert to hypertable. |
-| `time_column_name` | REGCLASS | Name of the column containing time values as well as the primary column to partition by. |
+| `time_column` | REGCLASS | Name of the column containing time values as well as the primary column to partition by. |
```
davidkohn88, line 20: This should not be changed, the parameter name is `time_column_name`. 

The datatype is wrong however, it's a `NAME` not a `REGCLASS`.

davidkohn88: Same with partitioning column below- should also technically be a name and not a regclass

davidkohn88: And partitioning_func / time_partitioning_func -> `REGPROC`
associated_schema_name -> `NAME`

davidkohn88 (commented):
Much better a few more things to help clarify here and a few odd errors some of which appear to be old.

timescaledb/how-to-guides/hypertables/create.md
charislam: Ohh good idea, the "key or constraint" thing is getting repetitive

charislam: ➕ I also think it is confusing and I've been struggling with how best to word it. Are the latest edits any clearer?

charislam: @davidkohn88 friendly bump to this 😃 for when you have a moment

api/create_hypertable.md
```diff
@@ -17,22 +17,22 @@ still work on the resulting hypertable.
 |Name|Type|Description|
 |---|---|---|
 | `relation` | REGCLASS | Identifier of table to convert to hypertable. |
-| `time_column_name` | REGCLASS | Name of the column containing time values as well as the primary column to partition by. |
+| `time_column_name` | NAME | Name of the column containing time values as well as the primary column to partition by. |
```
Loquacity, line 20: Please compress/unspace these tables.

charislam: Spun off to #908 to avoid cluttering review on this PR

charislam:
@davidkohn88 rewritten as we discussed!

api/create_distributed_hypertable.md
```diff
@@ -95,4 +95,16 @@ be removed in a future release. To avoid such inconsistency, we do not yet
...
+## Troubleshooting
+If you create a distributed hypertable on a table that already has a unique
+index or primary key, you might get this error:
+```
+ ERROR: cannot create a unique index without the column "<COLUMN_NAME>" (used in
+partitioning) 
+```
+
+To learn more and fix the problem, see the section on [hypertables and unique
+indexes][unique-indexes].
```
Loquacity, line 107:
```suggestion
For more information on how to fix this problem, see the
[hypertables and unique indexes][unique-indexes] section.
```

api/create_distributed_hypertable.md
```diff
@@ -95,4 +95,16 @@ be removed in a future release. To avoid such inconsistency, we do not yet
 recommend using `replication_factor` > 1, and instead rely on physical
 replication of each data node if such fault-tolerance is required.
 
+## Troubleshooting
```
Loquacity, line 98: Generally speaking, I prefer to have Troubleshooting as its own section, rather than tacked on the bottom of other content. Is that possible here? See timescaledb/how-to-guides/continuous-aggregates/troubleshooting.md for an example.

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
```diff
@@ -0,0 +1,140 @@
+# Hypertable partitioning and unique indexes
+Hypertable partitioning introduces some restrictions on unique indexes.
+
+## Partitioning columns
+To determine what unique indexes you can create on a hypertable, first
+identify your partitioning columns.
+
+In TimescaleDB, the following columns are partitioning columns:
```
Loquacity, line 8:
```suggestion
In TimescaleDB, you can use these columns for partitioning:
```

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
```diff
@@ -0,0 +1,140 @@
+# Hypertable partitioning and unique indexes
+Hypertable partitioning introduces some restrictions on unique indexes.
+
+## Partitioning columns
+To determine what unique indexes you can create on a hypertable, first
+identify your partitioning columns.
```
Loquacity, line 6:
```suggestion
Before you begin, you need to determine what unique indexes you can create on a hypertable.
Do this by identifying your partitioning columns.
```

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
```diff
@@ -0,0 +1,140 @@
...
+<highlight type="note">
+In PostgreSQL, a primary key is a unique index with a `NOT NULL` constraint.
+Any information about unique indexes also applies to primary keys.
+</highlight>
+
+### Create unique indexes on a hypertable
+Create a unique index as you normally would in PostgreSQL, using `CREATE UNIQUE
+INDEX`. Make sure to include all partitioning columns in the index.
+
+For example, say you create a hypertable partitioned on `time` and
```
Loquacity, line 42:
```suggestion
For example, to create a hypertable partitioned on `time` and
```

Loquacity (changes requested):

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
davidkohn88: This is less clear, these are the partitioning columns of a hypertable, there's no option, we use them for partitioning under the hood.

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
davidkohn88: We're not demonstrating this here, we're doing some setup, I think "say you" better conveys what's happening.

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
```diff
@@ -0,0 +1,140 @@
...
+  'time',
+  partitioning_column => 'device_id',
+  number_partitions => 4
+);
+```
+
+You can create a unique index on `time` and `device_id`:
+```sql
+CREATE UNIQUE INDEX idx_time_deviceid
+ON hypertable_example(time, device_id);
```
davidkohn88, line 63: This should be device_id, time, just cause usually an index with device_id leading is much better and I'd prefer we show that in the docs more frequently than the other way.

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+You can create a unique index on `time`, `user_id`, and `device_id`:
+```sql
+CREATE UNIQUE INDEX idx_time_userid_deviceid
+ON hypertable_example(time, user_id, device_id);
```
davidkohn88, line 69: same here, I'd have time as the trailing column of the index, so user_id, device_id, time

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+But you cannot create a unique index without `device_id`, because `device_id` is
+a partitioning column. This does not work:
+```sql
+-- This gives you an error
+CREATE UNIQUE INDEX idx_time
```
davidkohn88, line 76: I'd do this as a unique index on, say device_id, or user_id, device_id both because they're more realistic, and because I think it's important to demonstrate that the time column must be included in any unique index.

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
```diff
@@ -0,0 +1,140 @@
...
+-->
+
+If you don't include all the partitioning columns, you get this error:
+```
+ ERROR: cannot create a unique index without the column "<COLUMN_NAME>" (used in
+partitioning) 
+```
+
+<highlight type="note">
+In PostgreSQL, a primary key is a unique index with a `NOT NULL` constraint.
```
davidkohn88, line 34: I like the note, I think I'd put it higher up!

davidkohn88 (commented):
Looks much better! I think we're very close!

api/create_hypertable.md
```diff
@@ -232,3 +232,5 @@ important to create space partitioning, see
 [create_distributed_hypertable](/distributed-hypertables/create_distributed_hypertable/).
 The rare cases in which space partitions may be useful for non-distributed
 hypertables are described in the [add_dimension](/hypertable/add_dimension/) section.
+
+[create]: timescaledb/:currentVersion:/how-to-guides/hypertables/create/#hypertable-partitioning-with-unique-constraints
```
charislam, line 236:
```suggestion
```

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
charislam:
```suggestion
Before you create a unique index, you need to determine what unique indexes are allowed on your
hypertable. Begin by identifying your partitioning columns.
```

timescaledb/how-to-guides/hypertables/hypertables-and-unique-indexes.md
Loquacity:
```suggestion
TimescaleDB uses these columns for partitioning hypertables:
```

davidkohn88: That's fine with me.

davidkohn88 (approved):
This is looking good to me now!

Loquacity (approved):
LGTM, but my earlier comment about troubleshooting still stands.

charislam:
> LGTM, but my earlier comment about troubleshooting still stands.

@Loquacity, yes, I adopted it! There is a troubleshooting page now instead of separate troubleshooting sections :)

PR #805, opened 2022-02-10
Add linting rules

charislam:
# Description

Various linting rule improvements:

- Use em-dashes judiciously
- Standardize spellings and file extensions
- Make changes to word-substitution list
- Prevent flagging of parentheses in Markdown links
- Add reminder to replace `FIXME`s

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

### DIFF
--- a/install/installation-ubuntu-ami.md
+++ b/install/installation-ubuntu-ami.md
@@ -6,6 +6,8 @@ rather than the default disk that comes with EC2 instances.
 
 See below for the image id corresponding to each region for the most recent TimescaleDB version:
 
+<!-- vale Google.Units = NO -->
+
 Region | Image ID
 --- | ---
 us-east-1 (North Virginia) | ami-049af5874b750603f
@@ -17,6 +19,8 @@ eu-north-1 (Sweden) | ami-0d1b4497e3de9c999
 eu-west-1 (Ireland) | ami-016bd50ec09475098
 eu-west-2 (England) | ami-0983a20d47b9427c3
 eu-west-3 (France) | ami-00be4bb77bf15fe6a
+
+<!-- vale Google.Units = YES -->
 
 To launch the AMI, go to the `AMIs` section of your AWS EC2 Dashboard run the following steps:
 


.github/styles/Google/EmDashUse.yml
```diff
@@ -0,0 +1,7 @@
+extends: existence
+message: "Use em-dashes judiciously."
+link: 'https://developers.google.com/style/parentheses'
```
charislam, line 3:
```suggestion
```

Loquacity:
I like where this is going so far!

.github/styles/Google/WordList.yml
```diff
@@ -64,6 +69,7 @@ swap:
   open-source: open source
   overview screen: recents screen
```
Rajakavitha1, line 70: is this the name of the screen?

.github/styles/Google/WordList.yml
```diff
@@ -75,5 +81,7 @@ swap:
   tablet: device
   touch: tap
```
Rajakavitha1, line 82: in that case should tap & hold remain as is and not touch & hold?

charislam: Double-checked Google's style guide and they prefer 'touch & hold' (with the ampersand). I can edit this so it doesn't catch "touch & hold" (...but do we ever use the term? we don't have any mobile-first docs, do we?)

.github/styles/Google/WordList.yml
charislam: Good catch! This may be a purely Google thing. Checked Cloud Console and MST and did not find anything called 'recents'. There is an 'overview' though. Would it be properly be called an "overview pane" or an "overview tab"? @Loquacity 

image

charislam:
Modified parenthesis existence rules because I got tired of Vale flagging Markdown links for having parentheses. Here's the regex test:

image

.github/styles/Google/WordList.yml
Loquacity: I think that's a Google thing, it's not related to our products. As for the second question, whoever takes on reorganising the Cloud guide will need to make some decisions there (probably in conjunction with the Design team).

.github/styles/Google/WordList.yml
Loquacity: no, this doesn't apply to us.

Rajakavitha1 (approved):

Loquacity (approved):

PR #858, opened 2022-02-23
Update Windows release links

### LINKED ISSUE #725: There should be a complete list of latest version for all supported versions of PG
Refer to this thread: https://timescaledb.slack.com/archives/C4GT3N90X/p1641942539153400

Earlier you would first select your OS (Windows), and then your version of PG. Today, you're only linked to latest version of PG (today it's 14) even though the page says "Windows releases page". Seems the "Windows releases page" was never really created.

https://docs.timescale.com/install/latest/self-hosted/installation-windows/#install-self-hosted-timescaledb-on-windows-systems

charislam:
# Description

- Make latest release links evergreen
- Add links for PG12 and PG13

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Fixes #725

### DIFF
--- a/install/installation-windows.md
+++ b/install/installation-windows.md
@@ -135,4 +135,4 @@ if you want to have a chat.
 [tutorials]: /timescaledb/:currentVersion:/tutorials/
 [ms-download]: https://www.microsoft.com/en-us/download/details.aspx?id=48145
 [pg-download]: https://www.postgresql.org/download/windows/
-[windows-releases]: https://timescalereleases.blob.core.windows.net/windows/timescaledb-postgresql-14_2.5.1-windows-amd64.zip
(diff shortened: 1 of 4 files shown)


charislam:
> Hmm I dont think that fix is enough cause the page suggests there is a selection but this will get you pg14 binary directly. What we actually need is additional links for pg12 and pg13 too

Yeah actually there's an issue open for that problem: https://github.com/timescale/docs/issues/725
Do we have a Windows release page elsewhere? (the same way we have our Ubuntu releases on package cloud for example)
Or should I just create a docs page with those three links?

@svenklemm

mirandaauhl (approved):

PR #969, opened 2022-03-22
Clarify info about WITH NO DATA parameter

### LINKED ISSUE #845: [Docs RFC] Clarify caggs content about refreshing materialized view
# Describe change in content, appearance, or functionality

> I think the documentation is adding to the confusion. At this link: 
> https://docs.timescale.com/api/latest/continuous-aggregates/create_materialized_view/#create-materialized-view-continuous-aggregate
> 
> It says: 
> "The continuous aggregate view is automatically refreshed unless WITH NO DATA is given. This setting defaults to WITH DATA. For more information, see refresh_continuous_aggregate."
> 
> My reading of this is that the view will continue to be automatically refreshed unless I create it specifying WITH NO DATA. At first glance this appears to be happening because the created continuous aggregate has data when you do not specify “WITH NO DATA”. 
> 
> That said, it sounds like you are saying that ‘automatically refreshed’ in the documentation means that data is added on creation but not automatically updated over time? 
> 

> Agree with Jacob these docs are a little confusing.  I think it means to say:  "The continuous aggregate view is initially computed over all the current data in its underlying hypertable unless WITH NO DATA ..."  or something like that.
> 
> (My understanding is you will do a full pass over all the data when created if there isn't a "WITH NO DATA".  But it's a one-and-done.  If you want to continually do it, need a refresh plicy.)

# Subject matter expert (SME)

[If known, who is a good person to ask about this topic]

# Deadline

[When does this need to be addressed]

# Any further info

[Anything else you want to add, or further links]

Loquacity:
# Description

Rephrase section about WITH NO DATA parameter

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Fixes https://github.com/timescale/docs/issues/845


api/create_materialized_view.md
```diff
@@ -22,8 +22,12 @@ GROUP BY time_bucket( <const_value>, <partition_col_of_hypertable> ),
 [HAVING ...]
 ```
 
-The continuous aggregate view is automatically refreshed unless `WITH NO DATA`
-is given. This setting defaults to `WITH DATA`. For more information, see
+The continuous aggregate view defaults to `WITH DATA`. This means that when the
+view is created, it refreshes using all the current data in its underlying
+hypertable. This occurs once when the view is created. If you want the view to
+be refreshed regularly, you can use a refresh policy. If you do not want the
+view to update at all, even when it is first created, use the `WITH NO DATA`
```
charislam, line 29:
```suggestion
view to update when it is first created, use the `WITH NO DATA`
```

From my understanding, it would still update according to the refresh policy if you set one

charislam (approved):

Issue #1046, opened 2022-04-28
Document troubleshooting for Toolkit installation


charislam:
# Describe change in content, appearance, or functionality

Improve troubleshooting instructions for Toolkit installation.

> is the Toolkit package installed in the system properly? Can you prove it?
if so, is it listed in Postgres configuration to be loaded, the same way as the TimescaleDB extension certainly is?
if answers to the two previous questions are "yes", have you already restarted the Postgres?
and, if all conditions to have the Toolkit hold, do you see the extension in available extensions (the SQL query)?

[Relevant Slack thread](https://timescaledb.slack.com/archives/C4GT3N90X/p1651139232593579)

# Subject matter expert (SME)

[If known, who is a good person to ask about this topic]

# Deadline

[When does this need to be addressed]

# Any further info

[Anything else you want to add, or further links]

PR #1049, opened 2022-04-28
Remove extra extensions from install list

charislam:
# Description

Install instructions might be confusing because they list Toolkit even though Toolkit is not installed by default. Prune down to only the minimum extensions that you get if you follow the instructions.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links

Partial #1046


Rajakavitha1 (approved):
Yes, I had noticed this as well. We were planning on an IA for the redundant section in the entire install guide to be updated in one go. "Check the extension" section. There is inconsistency even in the name of the db that we create some section read `example` and our default extension is `tsdb`.

PR #1056, opened 2022-05-02
 Update installation instructions for Red Hat-based systems.

epgts:
# Description

 Update installation instructions for Red Hat-based systems.

- Use `yum` rather than `dnf` as it works on all systems documented
  herein, while `dnf` only exists on some.

- Document how to initialize database and enable and start
  postgresql service.

- Use `su postgres` to connect rather than postgres authentication as
  the latter is not configured by simply following our instructions and
  is beyond the scope of this document.

- Minimize references to systems other than CentOS 7 in Toolkit
  documentation, as that is the only system we test.

- Remove contradictory note telling user to "make sure you have
  installed PostgreSQL" right before telling them to "remove any
  existing PostgreSQL installations".

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links


charislam (approved):

Loquacity (approved):

mkindahl (approved):

epgts:
Ready to merge? Next up is Debian-based systems :)

Thanks!

PR #1079, opened 2022-05-09
Remove type parameter from frequency APIs

WireBaron:
# Description

In toolkit 1.7 the topn and frequency aggregates will no longer use PostgreSQL `AnyElement` types, but instead implicitly convert data to either integer or text formats.  This greatly simplifies the APIs for these aggregates by no longer requiring a type parameter to methods that need to return typed data.

The documentation should reflect the new API once toolkit 1.7 is released.

# Version

- [x] Latest (Default)
- [ ] Version 1.7
- [ ] Older [specify]

# Links


charislam (approved):

Loquacity:
Do we have an ETA for the release?

PR #1154, opened 2022-06-02
Add info about compression and caggs

charislam:
# Description

Some readers were missing the sections that spelled out the limitations
of compressed caggs. This adds the info explicitly to the introduction
of the cagg compression guide, also adds it to the troubleshooting
section, and makes some minor edits to the prior warning.

# Review checklists

- [ ] Is the content technically accurate?
- [ ] Is the content complete?
- [ ] Is the content presented in a logical order?
- [ ] Does the content use appropriate names for features and products?
- [ ] Does the content provide relevant links to further information?

- [x] Is the content free from typos?
- [x] Does the content use plain English?
- [x] Does the content contain clear sections for concepts, tasks, and references?
- [x] Are procedure and highlight tags used appropriately?
- [x] Has the index been updated appropriately?
- [x] Have any images been uploaded to the correct location, and are resolvable?
- [x] Are all links provided in reference style, and resolvable?
- [x] If the page index was updated, are redirects required
- [x] Have you checked the built version of this content?

### DIFF
--- a/timescaledb/how-to-guides/continuous-aggregates/compression-on-continuous-aggregates.md
+++ b/timescaledb/how-to-guides/continuous-aggregates/compression-on-continuous-aggregates.md
@@ -1,15 +1,21 @@
-# Compression on continuous aggregates
-Continuous aggregates are often used to store downsampled historical data.
-The historical data is almost never modified or recomputed and is only used 
-for serving analytic queries. For this use case, it is often beneficial to 
-store the materialized data in compressed form to save on storage costs. 
-You can get these cost savings by enabling compression on continuous 
-aggregates.
+# Compress continuous aggregates
+Continuous aggregates are often used to downsample historical data. If the data
+is only used for analytical queries and never modified, you can compress the
+aggregate to save on storage.
+
+Compression on continuous aggregates works similarly to
+[compression on hypertables][compression]. Compressed chunks can be queried, but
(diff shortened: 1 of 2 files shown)


Loquacity (approved):

NunoFilipeSantos:
I read at least 3 different portions of the text where we say we can't refresh a compressed cagg. Shouldn't we say it only once (list it as a limitation)?

charislam:
> I read at least 3 different portions of the text where we say we can't refresh a compressed cagg. Shouldn't we say it only once (list it as a limitation)?

In some cases, I might agree, but people were obviously missing the one place we put it before! (Another perfect example of callout blindness 😛 ) So we put it in the main text _and_ a callout because people will likely only read one or the other. And there's an associated error, so that should go in troubleshooting, where we list all the error text for people to search for.

Might create a bit of maintenance difficulty for us (we are working on ways to reduce that across the board with content reuse on the back end). But on the front end, it won't really be repetitive for readers, if that is your concern, because they are going to read 1, at most 2, of these.

charislam:
> haven't we started using reusable clips that we can quote on any part of the docs that automagically update?

Yep we have! I considered the best approach again based on your feedback and ended up taking out one of the repetitions. So now there are only two places where this info shows up (once in troubleshooting and once in the main text, where it's split between a callout and its adjacent text). Going to merge this PR for now so we can get the info on the site, but open to further refinement

PR #1199, opened 2022-06-16
Update hyperloglog with bucket guidance

WireBaron:
Update the hyperloglog documentation to recommend a minimum bucket
size of 1024.  Also update all examples to use larger bucket sizes.

Fixes https://github.com/timescale/product/issues/236

### DIFF
--- a/api/hyperloglog.md
+++ b/api/hyperloglog.md
@@ -22,11 +22,13 @@ For more information about approximate count distinct functions, see the
 
 |Name|Type|Description|
 |-|-|-|
-|buckets|integer|Number of buckets in the digest. Rounded up to the next power of 2, must be between 16 and 2^18.|
+|buckets|integer|Number of buckets in the digest. Rounded up to the next power of 2. Must be between 16 and 2^18, values less than 1024 are not recommended.|
 |value|AnyElement| Column to count distinct elements. The type must have an extended, 64-bit, hash function.|
 
 Increasing the `buckets` argument usually provides more accuracy at the expense
-of more storage.
+of more storage.  Because hyperloglog is a probabilistic algorithm, it works
+best on datasets that have many distinct values: at least tens of thousands. But it should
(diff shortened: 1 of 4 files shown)


Loquacity (approved):

api/hyperloglog.md
```diff
@@ -22,11 +22,13 @@ For more information about approximate count distinct functions, see the
 
 |Name|Type|Description|
 |-|-|-|
-|buckets|integer|Number of buckets in the digest. Rounded up to the next power of 2, must be between 16 and 2^18.|
+|buckets|integer|Number of buckets in the digest. Rounded up to the next power of 2, must be between 16 and 2^18, values less than 1024 are not recommended.|
```
charislam, line 25:
```suggestion
|buckets|integer|Number of buckets in the digest. Rounded up to the next power of 2. Must be between 16 and 2^18, values less than 1024 are not recommended.|
```

api/hyperloglog.md
```diff
@@ -22,11 +22,13 @@ For more information about approximate count distinct functions, see the
...
 |value|AnyElement| Column to count distinct elements. The type must have an extended, 64-bit, hash function.|
 
 Increasing the `buckets` argument usually provides more accuracy at the expense
-of more storage.
+of more storage.  Note that as a probabilistic algorithm, hyperloglog will work
+best on datasets with cardinality in at least the tens of thousands, but should
+also be fairly accurate when the number of buckets exceeds the cardinality.
```
charislam, line 31:
```suggestion
of more storage.  Because hyperloglog is a probabilistic algorithm, it works
best on datasets that have many distinct values: at least tens of thousands. But it should
be fairly accurate so long as you have more buckets than distinct values.
```

charislam (approved):
Nonblocking nitpicks

JLockerman (approved):

PR #1256, opened 2022-07-04
[Actions] Clean up and secure GitHub Actions

charislam:
Clean up and improve security of GitHub Actions.

- Pin all actions to an SHA (immutable) instead of a mutable tag or
  branch name.
- Add a Dependabot configuration file to keep pinned actions up to date.
  (Requires turning on Dependabot version updates for the repo. Is it OK
  if I do this, @Loquacity?)
- Explicitly define permissions for each workflow.
- Remove the PR creation workflow, which seems extraneous. (It puts a
  preview link on all PRs, even ones from forks. But previews from
  external forks don't get built.) Move the preview-link comment to the
  deploy workflow, which only deploys internal branches.
- Set Vale workflow to run only on PRs from internal branches. No point
  in having it try to run on external branches, as those don't have
  access to the required secret.
- Remove update of search index, which always fails due to server error.
  Search index is now updated as part of the web-documentation build
  process.


Loquacity:
> * Add a Dependabot configuration file to keep pinned actions up to date.
>   (Requires turning on Dependabot version updates for the repo. Is it OK
>   if I do this, @Loquacity?)

Go right ahead 👍

PR #1357, opened 2022-07-26
Configure markdownlint

charislam:
# Description

Adds an automatic formatter that can be used within your IDE. This is a complement to Vale, meant mainly to catch formatting inconsistencies and errors that can break build. Whereas Vale is meant mainly to catch style guide violations. There's some overlap between the two, with what goes where partially based on (1) what's easier to implement and (2) what's auto-fixable. (Markdownlint has auto-fixing but Vale doesn't seem to anymore(?) since Vale Server is deprecated.)

The chosen formatter is [markdownlint](https://github.com/DavidAnson/markdownlint) because it is broadly compatible. There is a VSCode extension, a Sublime Text extension, and a Vim/Neovim extension. You can also run it from the command line with node.js. This set-up has been tested with the VSCode extension.

## Required setup (VSCode)

- Installing the `markdownlint` extension.
- (optional but highly recommended) Configuring VS Code to auto-format on save by adding this to your `settings.json`: `"editor.formatOnSave": true`

## Use (VSCode)

With the above instructions, the extension will fix any errors that it can on save. Some errors aren't fixable because they require human review. Those will be highlighted with a colored squiggle and tooltip.

### Default rules
The default rules are configured for our preferred style. Here's a non-comprehensive list. I skipped some rules that never affect us, but that I didn't turn off as there's no real harm in keeping them on:

1. Headings should only increment one level at a time.
2. Headings should be in ATX style (hashes, not ASCII-doc type underlines).
3. Unordered lists should use an asterisk.
4. List items at the same level should be consistently indented.
5. Unordered lists should be indented by 4 spaces.
6. There should be no trailing whitespace.
7. There should be no hard tabs.
8. There should not be multiple consecutive blank lines.
9. Line length should cut off at 80 characters, with the exception of tables and code blocks.
10. Dollar signs should not be used before commands in code blocks if no output is shown.
11. There should be one and only one space after the last hash on a heading.
12. Headings should be surrounded by blank lines. (Currently we don't put a blank line between the heading and the following text, but I actually propose that we keep this rule on, as the lack of a blank line can mess up rewrapping, e.g., Vim's native rewrapper will consider the heading part of the following paragraph.)
13. Headings must start at the beginning of a line (no indent).
14. Multiple headings should not have the exact same content.
15. There should not be multiple top-level headings in a document.
16. Headings should not end in a period.
17. All ordered-list items should start with `1.`.
18. There should be 2 spaces after ordered-list markers and 3 spaces after unordered-list markers. That corresponds to an indentation of 4 spaces.
19. Fenced code blocks should be surrounded by blank lines.
20. Lists should be surrounded by blank lines.
21. Bare URLs should not be used.
22. Emphasis should not be used in place of an actual heading.
23. Fenced code blocks should have a language specified.
24. Links should not be empty.
25. Code blocks should use fenced style. (This will mainly prevent the unintentional creation of code blocks via indentation, one of Markdown's more annoying features that has bitten us a few times.)
26. Link fragments should be valid.
27. Reference links should use a defined label, and link definitions should be used. (In testing, this rule does not seem to work, have yet to figure out why.)

### Custom rules

Custom rules added so far:
1. Ensure a blank line on either side of a `<procedure>` tag. (autofixed)
1. Ensure there is a blank line after import statements. (autofixed)
1. Ensure procedure tags are closed. (not autofixed as there's no way to determine where closing tag should go)
1. Ensure procedure tags are not indented. (autofixed)
1. Ensure terminal tags are consistently indented. (autofixed)
1. Ensure there is a blank line after frontmatter. (autofixed)
1. There should be a line break, but no blank line, between a highlight tag and its content. (autofixed)
1. Highlight type should be one of the allowed types (currently 'note', 'important', 'warning', 'cloud', and 'deprecation'). (not autofixed as there's no way to determine what the type should be)


Rajakavitha1:
@charislam  this is awesome!!!! amazing work.

Loquacity (approved):

PR #1418, opened 2022-08-08
Jacob/markdownlint

jacobprall:
I mostly focused on the utils file, everything else looked solid. Here are some notes!

- I removed an inner function definition and extracted it to the module level. This is a stylistic choice. When working with function definitions, I prefer to place them on the outermost scope as necessary. Their placement, then, can be information - the function does not rely on any state or value outside itself. The "inner" function now has no access to the "outer" functions variables. Unless you're architecting an application with large functions that define multiple functions within themselves, I like to avoid nested function definitions if they aren't necessary. 
- One thing I noticed was some function calls with four unnamed variables. Common wisdom is to restrict a function's arguments to less than 3 - when you're passing 3 or more, there is a good chance they have a relationship and can be structured in an object. (as a side note, another reason is that the ordering of the parameters introduces unnecessary cognitive load to the next engineer, and a potential place for bugs to hide.) 
In this case, my suggestion is to pass the arguments as key/values in an object, and destructure them in the receiving function's function definition. This tells the IDE what the expected arguments are and helps with type-checking (yay autocomplete). While this pattern is basically default in React, I find it a helpful exercise in node, particularly when your list of parameters grows beyond two or three.


.vscode/markdownlint/utils.js
```diff
@@ -6,7 +6,7 @@
  * Adds a markdownlint error with a corresponding fix that inserts blank lines.
  *
  * @param {addErrorCallback} onError The callback that adds markdownlint errors.
- * @param {number} procedureLine 1-indexed line number of procedure tag.
+ * @param {number} lineNumber 1-indexed line number of procedure tag.
```
charislam, line 9:
```suggestion
 * @param {number} lineNumber 1-indexed line number to add the error on.
```

charislam (approved):

PR #1717, opened 2022-10-21
reorg(hyperfunctions): `stats_agg` functions

charislam:
Reorganize `stats_agg` functions: http://docs-dev.timescale.com/dev-feature-move-stats-aggs/api/dev-feature-move-stats-aggs/hyperfunctions/statistical-and-regression-analysis/stats_agg-one-variable/


Rajakavitha1 (approved):

Rajakavitha1:
Hi @charislam !!! I only looked through the deploy preview and all looks good to me.
A non blocking change maybe for later is to avoid using `foo` for table because it is it is not accessible compliant may be we can replace it with `example` or `TSDBs`.

PR #1781, opened 2022-11-17
fix(keywords)

charislam:
Go through keywords and make sure they return expected results. Remove some keywords that don't have good expected results because they are overly specific or vague.

Rename some other keywords for consistency: For templating reasons, the troubleshooting pages require the main topic (which is also a keyword) to be a noun, not a verb. Rather than creating yet another dictionary to map nouns to verbs, it is easier to just use nouns across the board.

### DIFF
--- a/install/index.md
+++ b/install/index.md
@@ -1,7 +1,7 @@
 ---
 title: Install TimescaleDB
 excerpt: Install TimescaleDB, the PostgreSQL database for time series and data analysis
-section: installation
+section: install
 nav-hidden: true
 keywords: [installation]
 ---
--- a/install/index.md
+++ b/install/index.md
@@ -1,9 +1,9 @@
 ---
 title: Install TimescaleDB
 excerpt: Install TimescaleDB, the PostgreSQL database for time series and data analysis
-section: install
+section: installation
 nav-hidden: true
-keywords: [install]
+keywords: [installation]
 ---
 
 # Install TimescaleDB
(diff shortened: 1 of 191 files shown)


install/index.md
```diff
@@ -1,9 +1,9 @@
 ---
 title: Install TimescaleDB
 excerpt: Install TimescaleDB, the PostgreSQL database for time series and data analysis
-section: install
+section: installation
```
charislam, line 4:
```suggestion
section: install
```

Loquacity:
Do we need to consider updating (or completely replacing) our keyword ontology?

charislam:
> Do we need to consider updating (or completely replacing) our keyword ontology?

Probably, I've been holding off because I want our metadata organization to be more stable before we officially update the ontology (to avoid continually overhauling it). It's not quite there yet because I still haven't touched tutorials/cloud/how-to-guide splash pages, so there are probably some needs I've overlooking right now

Shall we reevaluate in a few months?

Loquacity:
> > Do we need to consider updating (or completely replacing) our keyword ontology?
> 
> Probably, I've been holding off because I want our metadata organization to be more stable before we officially update the ontology (to avoid continually overhauling it). It's not quite there yet because I still haven't worked on functionality for tutorials/cloud/how-to-guide splash pages, so there are probably some needs I've overlooking right now
> 
> Shall we reevaluate in a few months?

Good plan. Let's look at it in the New Year.

Loquacity (approved):

PR #2329, opened 2023-05-02
Document next_start calculation on job failure

konskov:
# Description

# Links

# Writing help

# Review checklists

### DIFF
--- a/api/alter_job.md
+++ b/api/alter_job.md
@@ -90,3 +90,23 @@ Reschedules continuous aggregate job ID `1000` so that it next runs at 9:00:00 o
 ```sql
 SELECT alter_job(1000, next_start => '2020-03-15 09:00:00.0+00');
 ```
+
+### Calculation of next start on failure
+
+When a job run results in a runtime failure, the next start of the job is calculated taking into account both its `retry_period` and `schedule_interval`.
+The `next_start` time is calculated using the following formula:
+```
+next_start = finish_time + consecutive_failures * retry_period ± jitter
+```
+where jitter (± 13%) is added to avoid the "thundering herds" effect.
+
+<Highlight type="note">
+To ensure that the `next_start` time is not put off indefinitely or produce timestamps so large they end up out of range, it is capped at 5*`schedule_interval`.
+Also, more than 20 consecutive failures are not considered, so if the number of consecutive failures is higher, then it multiplies by 20.
+
+Additionally, in the case of jobs with fixed schedules, the system ensures that if the next start calculated as specified, surpasses the next scheduled execution, then the job is executed again at the next scheduled slot and not after that. This ensures that the job does not miss scheduled executions.
+
+Finally, there is a distinction between runtime failures that do not cause the job to crash and job crashes.
+In the event of a job crash, the next start calculation still follows the above formula,
+but it is always at least 5 minutes after the job's last finish, to give an operator enough time to disable it before another crash.
+</Highlight>


api/alter_job.md
```diff
@@ -90,3 +90,22 @@ Reschedules continuous aggregate job ID `1000` so that it next runs at 9:00:00 o
...
+
+<Highlight type="note">
+When a job run results in a runtime failure, the next start of the job is calculated taking into account both its `retry_period` and `schedule_interval`.
+The next_start is calculated using the following formula:
+```
+next_start = finish_time + consecutive_failures * retry_period
+```
+plus some jitter (± 13%) to avoid the “thundering herds” effect.
+
+However, as we don’t want to put off the next start indefinitely or produce timestamps so large they end up out of range, we cap this at 5* schedule_interval, and we also do not consider more than 20 consecutive failures, so if the number of consecutive failures is higher, we will multiply by 20 there.
```
Rajakavitha1, line 104:
```suggestion
However, to ensure that the `next start time` is not put off indefinitely or produce timestamps so large they end up out of range, it is capped at 5* `schedule_interval`. Also, more than 20 consecutive failures are not considered, so if the number of consecutive failures is higher, then it multiplies by 20.
```

api/alter_job.md
```diff
@@ -90,3 +90,22 @@ Reschedules continuous aggregate job ID `1000` so that it next runs at 9:00:00 o
...
+
+Additionally in the case of jobs with fixed schedules, we make sure that if the next start calculated as above, surpasses the next scheduled execution, then the job is executed again at the next scheduled slot and not after that, in order to ensure that the job does not miss scheduled executions.
```
Rajakavitha1, line 106:
```suggestion
Additionally, in the case of jobs with fixed schedules, the system ensures that if the next start calculated as specified, surpasses the next scheduled execution, then the job is executed again at the next scheduled slot and not after that. This ensures that the job does not miss scheduled executions.
```

api/alter_job.md
```diff
@@ -90,3 +90,22 @@ Reschedules continuous aggregate job ID `1000` so that it next runs at 9:00:00 o
...
+
+Finally, we distinguish between runtime failures that do not cause the job to crash and job crashes.
```
Rajakavitha1, line 108:
```suggestion
Finally, there is a distinction between runtime failures that do not cause the job to crash and job crashes.
```

api/alter_job.md
```diff
@@ -90,3 +90,22 @@ Reschedules continuous aggregate job ID `1000` so that it next runs at 9:00:00 o
...
+In the event of a job crash, the next start calculation still follows the above formula,
+but it will always be at least 5 minutes after the job's last finish, to give an operator enough time to disable it before another crash.
```
Rajakavitha1, line 110:
```suggestion
but it is always be at least 5 minutes after the job's last finish, to give an operator enough time to disable it before another crash.
```

Rajakavitha1 (changes requested):

api/alter_job.md
```diff
@@ -90,3 +90,22 @@ Reschedules continuous aggregate job ID `1000` so that it next runs at 9:00:00 o
 ```sql
 SELECT alter_job(1000, next_start => '2020-03-15 09:00:00.0+00');
 ```
+
+### Calculation of next start on failure
+
+<Highlight type="note">
+When a job run results in a runtime failure, the next start of the job is calculated taking into account both its `retry_period` and `schedule_interval`.
+The next_start is calculated using the following formula:
```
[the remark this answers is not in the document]
konskov: updated as requested

api/alter_job.md
konskov: updated as requested

Rajakavitha1 (approved):
Checked the build preview here: http://docs-dev.timescale.com/docs-konskov-document_failures_retry_period/api/konskov-document_failures_retry_period/actions/alter_job/

api/alter_job.md
```diff
@@ -90,3 +90,23 @@ Reschedules continuous aggregate job ID `1000` so that it next runs at 9:00:00 o
...
+### Calculation of next start on failure
+
+<Highlight type="note">
+When a job run results in a runtime failure, the next start of the job is calculated taking into account both its `retry_period` and `schedule_interval`.
+The `next_start` time is calculated using the following formula:
+```
+next_start = finish_time + consecutive_failures * retry_period
+```
+plus some jitter (± 13%) to avoid the “thundering herds” effect.
+
```
Rajakavitha1, line 103: <Highlight type="note">

api/alter_job.md
```diff
@@ -90,3 +90,23 @@ Reschedules continuous aggregate job ID `1000` so that it next runs at 9:00:00 o
...
+However, to ensure that the `next_start` time is not put off indefinitely or produce timestamps so large they end up out of range, it is capped at 5*`schedule_interval`.
```
Rajakavitha1, line 104:
```suggestion
To ensure that the `next_start` time is not put off indefinitely or produce timestamps so large they end up out of range, it is capped at 5*`schedule_interval`.
```

api/alter_job.md
Rajakavitha1, line 103: <Highlight type="note">

api/alter_job.md
```diff
@@ -90,3 +90,23 @@ Reschedules continuous aggregate job ID `1000` so that it next runs at 9:00:00 o
...
+
+### Calculation of next start on failure
+
+<Highlight type="note">
+When a job run results in a runtime failure, the next start of the job is calculated taking into account both its `retry_period` and `schedule_interval`.
+The `next_start` time is calculated using the following formula:
+```
+next_start = finish_time + consecutive_failures * retry_period
+```
+plus some jitter (± 13%) to avoid the “thundering herds” effect.
```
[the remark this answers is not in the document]
konskov: It feels to me that this phrasing implies that adding jitter is up to the user, when in fact is always added and the user can't control it. So, I wanted to confirm with you before making the suggested change whether that is the meaning we convey here or not?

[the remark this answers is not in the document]
Loquacity: Watch those smart quotes, too.

```suggestion
plus some jitter (± 13%) to avoid the "thundering herds" effect.
```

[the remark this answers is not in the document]
Rajakavitha1: > It feels to me that this phrasing implies that adding jitter is up to the user, when in fact is always added and the user can't control it. So, I wanted to confirm with you before making the suggested change whether that is the meaning we convey here or not?

In that case, maybe you want to add it as a part of the calculation formula.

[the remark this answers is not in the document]
konskov: I made a slight change there

Rajakavitha1 (approved):

PR #2372, opened 2023-05-15
GSG rework - Master PR

Loquacity:
# Description

Reworking the GSG

# Links

- Fixes: https://github.com/timescale/docs-private/issues/115

# Writing help

For information about style and word usage, see the [style guide](https://docs.timescale.com/about/latest/contribute-to-docs/)

# Review checklists

Reviewers: use this section to ensure you have checked everything before approving this PR:

## Subject matter expert (SME) review checklist

*   [ ] Is the content technically accurate?
*   [ ] Is the content complete?
*   [ ] Is the content presented in a logical order?
*   [ ] Does the content use appropriate names for features and products?
*   [ ] Does the content provide relevant links to further information?

## Documentation team review checklist

*   [ ] Is the content free from typos?
*   [ ] Does the content use plain English?
*   [ ] Does the content contain clear sections for concepts, tasks, and references?
*   [ ] Have any images been uploaded to the correct location, and are resolvable?
*   [ ] If the page index was updated, are redirects required
      and have they been implemented?
*   [ ] Have you checked the built version of this content?

PR #2427, opened 2023-06-07
feat: add frontmatter for curricular content

charislam:
Add frontmatter to group curricular content by tutorial, and to choose whether to use the highlighted next and previous buttons for those pages.

Related to https://github.com/timescale/web-documentation/pull/617
Related to https://github.com/timescale/docs-private/issues/100

### DIFF
--- a/getting-started/add-data.md
+++ b/getting-started/add-data.md
@@ -4,6 +4,8 @@ excerpt: Add time-series data to your Timescale instance
 products: [cloud, mst, self_hosted]
 keywords: [ingest]
 tags: [add, data, time-series]
+layout_components: [next_prev_large]
+content_group: Getting started
 ---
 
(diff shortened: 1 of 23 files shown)


Loquacity:
@charislam Am I correct in assuming we only need to add these to the ones we want to use the treatment on? So we don't need to update all pages?

charislam:
> @charislam Am I correct in assuming we only need to add this metadata to the ones we want to use the treatment on? So we don't need to update all pages?

Yep I think I already got all of the ones we need in this PR :D

Loquacity (approved):
Approved, with the note that I've added this frontmatter to the GSG rework in https://github.com/timescale/docs/pull/2372 but I should be able to manage the conflicts if this merges first.

PR #2664, opened 2023-08-30
Update Compression content

Loquacity:
# Description

Updates the existing Compression content, and includes more how to content.

# Links

Fixes https://github.com/timescale/docs-private/issues/167

# Writing help

# Review checklists

### DIFF
--- a/use-timescale/compression/compression-design.md
+++ b/use-timescale/compression/compression-design.md
@@ -7,14 +7,7 @@ keywords: [compression, schema, tables]
 
 # Designing for compression
 
-Traditionally, databases are considered either row-based, or column based. And
-each type of database brings benefits and drawbacks, including query speed,
-insert speed, and the level to which they can effectively compress data.
-Generally speaking, column-oriented databases are highly compressible, but
-inserting data can take longer. Conversely, row-oriented databases have faster
-queries, but can't compress as well.
-
-Time-series data can be unique, in that it needs to handle both shall and wide
+Time-series data can be unique, in that it needs to handle both shallow and wide
 queries, such as "What's happened across the deployment in the last 10 minutes,"
 and deep and narrow, such as "What is the average CPU usage for this server
 over the last 24 hours." Time-series data usually has a very high rate of
@@ -28,13 +21,13 @@ consider the design of your database, before you start ingesting data. This
 section covers some of the things you need to take into consideration when
 designing your database for maximum compression effectiveness.
 
-## Array format
+## Compressing data
 
 TimescaleDB is built on PostgreSQL which is, by nature, a row-based database.
-Because time-series data is accessed in order of time, TimescaleDB converts many
-wide rows of data into a single row of data, called an array form. This means
-that each field of that new, wide row stores an ordered set of data comprising
-the entire column.
+Because time-series data is accessed in order of time, when you enable
+compression, TimescaleDB converts many wide rows of data into a single row of
+data, called an array form. This means that each field of that new, wide row
+stores an ordered set of data comprising the entire column.
 
 For example, if you had a table with data that looked a bit like this:
 
@@ -54,11 +47,9 @@ You can convert this to a single row in array form, like this:
 |[12:00:01, 12:00:01, 12:00:02, 12:00:02, 12:00:03, 12:00:03]|[A, B, A, B, A, B]|[0, 0, 0, 0, 0, 4]|[70.11, 69.70, 70.12, 69.69, 70.14, 69.70]|
 
 Even before you compress any data, this format immediately saves storage by
-reducing the per-row overhead. PostgreSQL typically adds around 27 bytes of
-overhead per row. So even without any compression, if our schema above is say 32
-bytes, then 1000 rows of data which previously took about 59 kilobytes
-(`1000 x (32 + 27) ~= 59 kilobytes`), now takes about 32 kilobytes
-(`1000 x 32 + 27 ~= 32 kilobytes`) in this format.
+reducing the per-row overhead. PostgreSQL typically adds a small number of bytes
+of overhead per row. So even without any compression, the schema in this example
+is now smaller on disk than the previous format.
 
 This format arranges the data so that similar data, such as timestamps, device
 IDs, or temperature readings, is stored contiguously. This means that you can
@@ -100,12 +91,13 @@ required.
 
 TimescaleDB automatically includes more information in the row and includes
 additional groupings to improve query performance. When you compress a
-hypertable, either manually or through a compression policy, you need to specify
+hypertable, either manually or through a compression policy, it can help to specify
 an `ORDER BY` column.
 
-`ORDER BY` columns specify how the rows that are part of a compressed patch are
-ordered. For most time-series workloads, this is by timestamp, but you can also
-specify a second dimension, such as location.
+`ORDER BY` columns specify how the rows that are part of a compressed batch are
+ordered. For most time-series workloads, this is by timestamp, so if you don't
+specify an `ORDER BY` column, TimescaleDB defaults to using the time column. You
+can also specify additional dimensions, such as location.
 
 For each `ORDER BY` column, TimescaleDB automatically creates additional columns
 that store the minimum and maximum value of that column. This way, the query
@@ -114,10 +106,11 @@ having to do any decompression, and determine whether the row could possibly
 match the query.
 
 When you compress your hypertable, you can also choose to specify a `SEGMENT BY`
-column. This allows you to segment compressed rows by a specific column,
-so that each compressed row corresponds to a data about a single item such as,
-for example, a specific device ID. This further allows the query planner to
-determine if the row could possibly match the query. For example:
+column. This allows you to segment compressed rows by a specific column, so that
+each compressed row corresponds to a data about a single item such as, for
+example, a specific device ID. This further allows the query planner to
+determine if the row could possibly match the query without having to decompress
+the column first. For example:
 
 |Device ID|Timestamp|Status Code|Temperature|Min Timestamp|Max Timestamp|
 |-|-|-|-|-|-|
--- a/use-timescale/compression/compression-design.md
+++ b/use-timescale/compression/compression-design.md
@@ -121,7 +121,7 @@ With the data segmented in this way, a query for device A between a time
 interval becomes quite fast. The query planner can use an index to find those
 rows for device A that contain at least some timestamps corresponding to the
 specified interval, and even a sequential scan is quite fast since evaluating
-device IDs or timestamps does not require decompression. This means the the
+device IDs or timestamps does not require decompression. This means the
 query executor only decompresses the timestamp and temperature columns
 corresponding to those selected rows.
 
--- a/use-timescale/compression/compression-design.md
+++ b/use-timescale/compression/compression-design.md
@@ -1,0 +1,135 @@
+---
+title: Designing your database for compression
+excerpt: Learn how to design your database for the most effective compression
+products: [cloud, mst, self_hosted]
+keywords: [compression, schema, tables]
+---
+
+# Designing for compression
+
+Traditionally, databases are considered either row-based, or column based. And
+each type of database brings benefits and drawbacks, including query speed,
+insert speed, and the level to which they can effectively compress data.
+Generally speaking, column-oriented databases are highly compressible, but
+inserting data can take longer. Conversely, row-oriented databases have faster
+queries, but can't compress as well.
+
+Time-series data can be unique, in that it needs to handle both shall and wide
+queries, such as "What's happened across the deployment in the last 10 minutes,"
+and deep and narrow, such as "What is the average CPU usage for this server
+over the last 24 hours." Time-series data usually has a very high rate of
+inserts as well; hundreds of thousands of writes per second can be very normal
+for a time-series dataset. Additionally, time-series data is often very
+granular, and data is collected at a higher resolution than many other
+datasets. This can result in terabytes of data being collected over time.
+
+All this means that if you need great compression rates, you probably need to
+consider the design of your database, before you start ingesting data. This
+section covers some of the things you need to take into consideration when
+designing your database for maximum compression effectiveness.
+
+## Array format
+
+TimescaleDB is built on PostgreSQL which is, by nature, a row-based database.
+Because time-series data is accessed in order of time, TimescaleDB converts many
+wide rows of data into a single row of data, called an array form. This means
+that each field of that new, wide row stores an ordered set of data comprising
+the entire column.
+
+For example, if you had a table with data that looked a bit like this:
+
+|Timestamp|Device ID|Status Code|Temperature|
+|-|-|-|-|
+|12:00:01|A|0|70.11|
+|12:00:01|B|0|69.70|
+|12:00:02|A|0|70.12|
+|12:00:02|B|0|69.69|
+|12:00:03|A|0|70.14|
+|12:00:03|B|4|69.70|
+
+You can convert this to a single row in array form, like this:
+
+|Timestamp|Device ID|Status Code|Temperature|
+|-|-|-|-|
+|[12:00:01, 12:00:01, 12:00:02, 12:00:02, 12:00:03, 12:00:03]|[A, B, A, B, A, B]|[0, 0, 0, 0, 0, 4]|[70.11, 69.70, 70.12, 69.69, 70.14, 69.70]|
+
+Even before you compress any data, this format immediately saves storage by
+reducing the per-row overhead. PostgreSQL typically adds around 27 bytes of
+overhead per row. So even without any compression, if our schema above is say 32
+bytes, then 1000 rows of data which previously took about 59 kilobytes
+(`1000 x (32 + 27) ~= 59 kilobytes`), now takes about 32 kilobytes
+(`1000 x 32 + 27 ~= 32 kilobytes`) in this format.
+
+This format arranges the data so that similar data, such as timestamps, device
+IDs, or temperature readings, is stored contiguously. This means that you can
+then use type-specific compression algorithms to compress the data further, and
+each array is separately compressed. For more information about the compression
+methods used, see the [compression methods section][compression-methods].
+
+When the data is in array format, you can perform queries that require a subset
+of the columns very quickly. For example, if you have a query like this one, that
+asks for the average temperature over the past day:
+
+<CodeBlock canCopy={false} showLineNumbers={false} children={`
+SELECT time_bucket(‘1 minute’, timestamp) as minute
+ AVG(temperature)
+FROM table
+WHERE timestamp > now() - interval ‘1 day’
+ORDER BY minute DESC
+GROUP BY minute;
+`} />
+
+The query engine can fetch and decompress only the timestamp and temperature
+columns to efficiently compute and return these results.
+
+Finally, TimescaleDB uses non-inline disk pages to store the compressed arrays.
+This means that the in-row data points to a secondary disk page that stores the
+compressed array, and the actual row in the main table becomes very small,
+because it is now just pointers to the data. When data stored like this is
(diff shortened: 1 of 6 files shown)


Loquacity:
@iroussos This is still in draft, but adding you so you can follow along as I go, and provide advice 💖

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,135 @@
...
+required.
+
+TimescaleDB automatically includes more information in the row and includes
+additional groupings to improve query performance. When you compress a
+hypertable, either manually or through a compression policy, you need to specify
+an `ORDER BY` column.
+
+`ORDER BY` columns specify how the rows that are part of a compressed patch are
+ordered. For most time-series workloads, this is by timestamp, but you can also
+specify a second dimension, such as location.
```
mkindahl, line 108: You can specify as many columns as you like, even though it might not be very efficient. This sounds like you can only add one extra column.

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,135 @@
...
+that store the minimum and maximum value of that column. This way, the query
+planner can look at the range of timestamps in the compressed column, without
+having to do any decompression, and determine whether the row could possibly
+match the query.
+
+When you compress your hypertable, you can also choose to specify a `SEGMENT BY`
+column. This allows you to segment compressed rows by a specific column,
+so that each compressed row corresponds to a data about a single item such as,
+for example, a specific device ID. This further allows the query planner to
+determine if the row could possibly match the query. For example:
```
mkindahl, line 120:
```suggestion
for example, a specific device ID. This further allows the query planner to
determine if the row could possibly match the query without having to decompress the column. For example:
```

mkindahl (commented):
Reviewing section "about compression". Will continue with the others.

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,135 @@
...
+FROM table
+WHERE timestamp > now() - interval ‘1 day’
+ORDER BY minute DESC
+GROUP BY minute;
+`} />
+
+The query engine can fetch and decompress only the timestamp and temperature
+columns to efficiently compute and return these results.
+
+Finally, TimescaleDB uses non-inline disk pages to store the compressed arrays.
```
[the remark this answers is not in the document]
Loquacity: Yeah, I was trying to avoid the TOAST stuff, because it was going to require a whole lot more explanation 😅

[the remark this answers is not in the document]
mkindahl: Seems reasonable.

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,135 @@
...
+This means that the in-row data points to a secondary disk page that stores the
+compressed array, and the actual row in the main table becomes very small,
+because it is now just pointers to the data. When data stored like this is
+queried, only the compressed arrays for the required columns are read from disk,
+further improving performance by reducing disk reads and writes.
+
+## Querying compressed data
```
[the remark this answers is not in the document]
Loquacity: I might need some more concrete examples from you here, if you're able ... ?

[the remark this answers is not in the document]
mkindahl: Here is a very rough blurb of text for the "order-by" example I suggested above.

> ## Ordering data in compressed columns
> 
> A very common way to compress data is to use delta encoding. The default
> algorithm for storing timestamps is the delta-delta algorithm, which will
> compress the data even more, but for this section the delta-encoding is used
> since it is easier to explain. Using delta encoding on the timestamp column will
> store the difference between consecutive values in the column, like this:
> 
> | Timestamp | Device ID | Status Code | Temperature |
> |-----------|-----------|-------------|-------------|
> | 12:00:01  | A         | 0           | 70.11       |
> | 1 second  | A         | 0           | 70.12       |
> | 1 second  | A         | 0           | 70.14       |
> | 12:00:01  | B         | 0           | 69.70       |
> | 1 second  | B         | 0           | 69.69       |
> | 1 second  | B         | 4           | 69.70       |
> 
> Since timestamps are stored at microsecond precision and are represented as a
> 64-bit integer, this would allow the timestamp column to be stored as 64 bits
> for the time and the increments using 20 bits for a total of 26 bytes rather
> than 48 bytes giving a compression of 46%.
> 
> If you looked carefully at the example above, you might have noted that the time
> column was not increasing in time. This can happen if you insert data into the
> hypertable in batches but it creates a problem for the compression
> algorithms. If the rows were ordered in timestamp order instead, you would be
> able to compress the data even more:
> 
> | Timestamp | Device ID | Status Code | Temperature |
> |-----------|-----------|-------------|-------------|
> | 12:00:01  | A         | 0           | 70.11       |
> | 0 seconds | B         | 0           | 69.70       |
> | 1 second  | A         | 0           | 70.12       |
> | 0 seconds | B         | 0           | 69.69       |
> | 1 second  | A         | 0           | 70.14       |
> | 0 seconds | B         | 4           | 69.70       |
> 
> This would compress the timestamps to 64 bits plus 2 times 20 bits, which is a
> whopping 73% compression!
> 
> TimescaleDB will automatically order the rows using the time column, but it also
> allows you to add additional columns to use for ordering using the
> [`compression_orderby` option][compression_orderby].
>

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,128 @@
...
+itself is in a compressed column. You don't want to have to decompress all the
+data in a chunk, or even an entire hypertable, to determine which rows are
+required.
+
+TimescaleDB automatically includes more information in the row and includes
+additional groupings to improve query performance. When you compress a
+hypertable, either manually or through a compression policy, it can help to specify
+an `ORDER BY` column.
+
+`ORDER BY` columns specify how the rows that are part of a compressed batch are
```
mkindahl, line 97:
```suggestion
order-by columns specify how the rows that are part of a compressed batch are
```

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,128 @@
...
+ordered. For most time-series workloads, this is by timestamp, so if you don't
+specify an `ORDER BY` column, TimescaleDB defaults to using the time column. You
```
mkindahl, line 99:
```suggestion
specify an order-by column, TimescaleDB defaults to using the time column. You
```

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,128 @@
...
+can also specify additional dimensions, such as location.
+
+For each `ORDER BY` column, TimescaleDB automatically creates additional columns
```
mkindahl, line 102:
```suggestion
For each order-by column, TimescaleDB automatically creates additional columns
```

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,128 @@
...
+that store the minimum and maximum value of that column. This way, the query
+planner can look at the range of timestamps in the compressed column, without
+having to do any decompression, and determine whether the row could possibly
+match the query.
+
+When you compress your hypertable, you can also choose to specify a `SEGMENT BY`
```
mkindahl, line 108: There is no `SEGMENT BY` in SQL, but this format suggest there is, which is a little confusing.
```suggestion
When you compress your hypertable, you can also choose to specify a segment-by
```

mkindahl (commented):
Some suggestions regarding use of `ORDER BY` and `SEGMENT BY`.

use-timescale/compression/compression-design.md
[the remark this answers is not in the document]
mkindahl: Here is a very rough version of the "query compressed data" that brings in a query on the segment-by It is basically using the existing section, but adds a query:

> ## Using segment-by to improve query performance for compressed data 
> 
> The compressed data in TimescaleDB are treated as opaque data by PostgreSQL,
> meaning that it is not possible to look at a compressed batch and know what
> information is in that batch. To be able to figure out if a batch needs to be
> compressed when reading it in a query, TimescaleDB automatically includes more
> information for each batch to improve query performance.
> 
> For order-by columns, TimescaleDB automatically adds additional data that stores
> the minimum and maximum value of that column. This way, the query planner can
> look at the range of timestamps in the compressed column, without having to
> decompress a batch, and determine whether the data in the batch could possibly
> match the query.
> 
> If you, however, want to query based on the device ID, like this, would have to
> decompress the device column to decide if that batch should be part of the
> query.
> 
> <CodeBlock canCopy={false} showLineNumbers={false} children={`
> SELECT time_bucket(‘1 minute’, timestamp) as minute
>  AVG(temperature)
> FROM table
> WHERE timestamp > now() - interval ‘1 day’
>   AND device_id = 'A'
> ORDER BY minute DESC;
> `} />
> 
> If you are often running queries like this, it would be better to not compress
> the device id column at all and instead create batches like this.
> 
> | Device ID | Timestamp                      | Status Code | Temperature           | Min Timestamp | Max Timestamp |
> |-----------|--------------------------------|-------------|-----------------------|---------------|---------------|
> | A         | [12:00:01, 12:00:02, 12:00:03] | [0, 0, 0]   | [70.11, 70.12, 70.14] | 12:00:01      | 12:00:03      |
> | B         | [12:00:01, 12:00:02, 12:00:03] | [0, 0, 0]   | [70.11, 70.12, 70.14] | 12:00:01      | 12:00:03      |
> 
> With the data segmented in this way, a query for device A between a time
> interval becomes quite fast. The query planner can use an index to find those
> rows for device A that contain at least some timestamps corresponding to the
> specified interval, and even a sequential scan is quite fast since evaluating
> device IDs or timestamps does not require decompression. This means the query
> executor only decompresses the timestamp and temperature columns corresponding
> to those selected rows.
> 
> Fortunately, TimescaleDB allow you to provide zero or more segment-by columns
> using the `compression_segmentby` option, which allows you to get good
> performance also for this case.
>

use-timescale/compression/compression-design.md
```diff
@@ -80,49 +80,86 @@ because it is now just pointers to the data. When data stored like this is
...
-specify an `ORDER BY` column, TimescaleDB defaults to using the time column. You
-can also specify additional dimensions, such as location.
-
-For each `ORDER BY` column, TimescaleDB automatically creates additional columns
-that store the minimum and maximum value of that column. This way, the query
-planner can look at the range of timestamps in the compressed column, without
-having to do any decompression, and determine whether the row could possibly
+## Ordering compressed data
+
+When data is compressed using [delta encoding][delta], the column being encoded
```
mkindahl, line 85:
```suggestion
The order of the data can make a significant difference to the compression rate that you can achieve. As an example, if data were compressed using [delta encoding][delta], the column being encoded
```

use-timescale/compression/compression-design.md
```diff
@@ -80,49 +80,86 @@ because it is now just pointers to the data. When data stored like this is
...
+|0 seconds|B|0|69.69|
+|1 second|A|0|70.14|
+|0 seconds|B|4|69.70|
+
+This compresses the timestamps to 64 bits plus 2 * 20 bits, which is 73%
+compression, and much higher than the first example, for exactly the same data.
+
+By default, TimescaleDB orders data by time, but you can also add additional
+columns for ordering.
+
```
mkindahl, line 126:
```suggestion
By default, TimescaleDB orders data by time, but you can also add additional
columns for ordering.

TimescaleDB uses a more advanced compression algorithm for timestamps called delta-delta encoding which gives even better compression ratio for timestamps, but the delta-encoding above was used for illustrative purposes.
```

use-timescale/compression/compression-design.md
```diff
@@ -80,49 +80,86 @@ because it is now just pointers to the data. When data stored like this is
...
-example, a specific device ID. This further allows the query planner to
-determine if the row could possibly match the query without having to decompress
-the column first. For example:
+However, using the previous example, if you wanted to query the data based on
+device ID, you would have to decompress the device column to decide if that
+batch should be part of the query. When you compress your hypertable, you can
+choose to specify a `SEGMENT BY` column. This allows you to segment compressed
+rows by a specific column, so that each compressed row corresponds to a data
+about a single item such as, for example, a specific device ID. For example, you
+could use `SEGMENT_BY` to create batches like this:
```
mkindahl, line 149: Consistent? Still not sure if you want to use `SEGMENT BY` or "segment-by" in the text.

```suggestion
could use `SEGMENT BY` to create batches like this:
```

use-timescale/compression/compression-design.md
```diff
@@ -80,49 +80,86 @@ because it is now just pointers to the data. When data stored like this is
...
 
 |Device ID|Timestamp|Status Code|Temperature|Min Timestamp|Max Timestamp|
 |-|-|-|-|-|-|
 |A|[12:00:01, 12:00:02, 12:00:03]|[0, 0, 0]|[70.11, 70.12, 70.14]|12:00:01|12:00:03|
 |B|[12:00:01, 12:00:02, 12:00:03]|[0, 0, 0]|[70.11, 70.12, 70.14]|12:00:01|12:00:03|
 
-With the data segmented in this way, a query for device A between a time
+With the data segmented this way, a query for device A during a particular time
```
mkindahl, line 156:
```suggestion
With the data segmented this way, a query for device A in a particular time
```

use-timescale/compression/compression-design.md
```diff
@@ -80,49 +80,86 @@ because it is now just pointers to the data. When data stored like this is
...
 interval becomes quite fast. The query planner can use an index to find those
 rows for device A that contain at least some timestamps corresponding to the
-specified interval, and even a sequential scan is quite fast since evaluating
-device IDs or timestamps does not require decompression. This means the
-query executor only decompresses the timestamp and temperature columns
-corresponding to those selected rows.
+specified interval, and even a sequential scan is quite fast, since evaluating
+device IDs or timestamps does not require decompression. This means the query
+executor only decompresses the timestamp and temperature columns corresponding
+to those selected rows.
```
mkindahl, line 162:
```suggestion
specified interval, and even a sequential scan is quite fast, since evaluating
device IDs or timestamps does not require decompression. This means the query
executor only decompresses the timestamp and temperature columns corresponding
to those selected rows.

This, however, comes at the expense of a worse compression ratio. If you look at the example above, you now have two rows with three items in each compressed column rather than a single row with six items in the compressed rows.
```

mkindahl (commented):
Some minor comments, otherwise it looks accurate. I wonder if you want to mention that compression stores at most 1000 elements in a compressed column somewhere. It affects the compression ratio if you have few elements for each segment-by column.

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,177 @@
...
+
+This compresses the timestamps to 64 bits plus 2 * 20 bits, which is 73%
+compression, and much higher than the first example, for exactly the same data.
+
+By default, TimescaleDB orders data by time, but you can also add additional
+columns for ordering.
+
+## Segmenting
+
+compressed data
```
mkindahl, line 134: Typo?
```suggestion
## Segmenting compressed data
```

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,177 @@
...
+
+When you have compressed your data in this way, the database has no way of
```
mkindahl, line 136:
```suggestion
When you have compressed your data in the way described in the previous section, the database has no way of
```

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,176 @@
...
+## Segmenting compressed data
+
+When you have compressed your data in the way described in the previous section,
+the database has no way of knowing which rows need to be fetched and
+decompressed to resolve a query. For example, the database can't easily
+determine which rows contain data from the past day, as the timestamp itself is
+in a compressed column. You don't want to have to decompress all the data in a
+chunk, or even an entire hypertable, to determine which rows are required.
+TimescaleDB automatically includes more information in each batch to improve
+query performance.
```
mkindahl, line 141:
```suggestion
To be able to determine if a row potentially can match a query, TimescaleDB automatically includes more information in each batch.
```

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,176 @@
...
+decompress a batch, and determine whether the data in the batch could possibly
+match the query.
+
+However, using the previous example, if you wanted to query the data based on
+device ID, you would have to decompress the device column to decide if that
+batch should be part of the query. When you compress your hypertable, you can
+choose to specify a `SEGMENT BY` column. This allows you to segment compressed
+rows by a specific column, so that each compressed row corresponds to a data
+about a single item such as, for example, a specific device ID. For example, you
+could use `SEGMENT BY` to create batches like this:
```
mkindahl, line 155: Not sure if this was intentional.

```suggestion
choose to specify a segment-by column. This allows you to segment compressed
rows by a specific column, so that each compressed row corresponds to a data
about a single item such as, for example, a specific device ID. For example, you
could use segment-by to create batches like this:
```

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,176 @@
...
+
+For order-by columns, TimescaleDB automatically adds additional data that stores
+the minimum and maximum value of that column. This way, the query planner can
+look at the range of timestamps in the compressed column, without having to
+decompress a batch, and determine whether the data in the batch could possibly
+match the query.
+
+However, using the previous example, if you wanted to query the data based on
+device ID, you would have to decompress the device column to decide if that
+batch should be part of the query. When you compress your hypertable, you can
```
mkindahl, line 151: It is not entirely clear why the entire column have to be decompressed.

One reason is because every compressed row is likely to contain every device, which means that having a max and a min does not help in deciding if the row need to be decompressed: the test indicates that the column always has to be decompressed.

The other reason is because we cannot have an index on data in a compressed column.

use-timescale/compression/compression-design.md
```diff
@@ -0,0 +1,176 @@
...
+interval becomes quite fast. The query planner can use an index to find those
+rows for device A that contain at least some timestamps corresponding to the
+specified interval, and even a sequential scan is quite fast, since evaluating
+device IDs or timestamps does not require decompression. This means the query
+executor only decompresses the timestamp and temperature columns corresponding
+to those selected rows.
+
+This, however, comes at the expense of a worse compression ratio. In the
+example, you now have two rows with three items in each compressed column rather
+than a single row with six items in the compressed rows.
```
mkindahl, line 172: This is only true if the batches contain less than 1000 values, so if each device has more than 1000 values inside a chunk, it should still compress well. You might want to elaborate on this a little more.

mkindahl (commented):
A few more suggestions for improvements.

use-timescale/compression/compression-design.md
Loquacity: My interpretation was the former: that you don't know if the device is referenced in the column without decompressing it first.

mkindahl: Sorry, I was unclear. Both reasons are valid, but let me elaborate a little.

Typically, device IDs are taken from a separate table containing a serial ID. This means that device IDs typically are integers, not strings. With this in mind, we can compress the device ID column as well, which gives us something like this:

| `device`              | `_ts_meta_min_1` | `_ts_meta_max_1` |
|-----------------------|------------------|------------------|
| [0, 1, 2, 1, 2, 3, 0] | 0                | 3                |
| [0, 0, 3, 0, 0, 2, 0] | 0                | 3                |

However, as you can see, the range does not allow us to filter out the rows that does not contain 1 since the entire range of device IDs are covered in each row, so we have to scan all rows and decompress the column to see if there is a match.

However, if we had an index on the device ID, we would be able to select all the compressed rows for that device, so by segmenting the data by the device ID, we can easily select the rows that have matching IDs using an index.

iroussos:
Reopening this PR as it seems to have been closed by accident

iroussos (approved):
Thank you @mkindahl, looks really good; happy to get this out and then keep on iterating!

PR #2666, opened 2023-09-01
WIP: pg_dump and pg_restore docs: Postgres source

alejandrodnm:
# Description

# Links

# Writing help

# Review checklists

### DIFF
--- a/use-timescale/migration/dual-write-and-backfill.md
+++ b/use-timescale/migration/dual-write-and-backfill.md
@@ -87,7 +87,12 @@ pg_dumpall -d "$SOURCE" \
   --file=roles.sql
 ```
 
-Fix up the dumped roles:
+Timescale services do not support roles with superuser access. If your SQL
+dump includes roles that have such permissions, you'll need to modify the file
+to be compliant with the security model.
+
+You can use the following `sed` command to remove unsupported statements and
+permissions from your roles.sql file:
 
 ```
 sed -i -E \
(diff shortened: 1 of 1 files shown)


use-timescale/migration/dual-write-and-backfill.md
```diff
@@ -157,9 +185,101 @@ commit;
...
+
+```
+psql -X -d "$TARGET" \
+  -v ON_ERROR_STOP=1 \
+  --echo-errors \
+  -f roles.sql \
+  -f dump.sql
+```
+
+TODO: Should we add a mention that customers can exclude the data from large
```
alejandrodnm, line 280: @jgpruitt WDYT?

use-timescale/migration/dual-write-and-backfill.md
```diff
@@ -157,9 +185,101 @@ commit;
...
+  --no-privileges \
+  --file=dump.sql
+```
+
+a. `--no-tablespaces` is required because Timescale does not support
+   tablespaces other than the default. This is a limitation.
+b. `--no-owner` is required because tsdbadmin is not a superuser and cannot
+   assign ownership in all cases. This flag means that everything will be
+   owned by the tsdbadmin user in the target regardless of ownership in the
+   source. This is a limitation.
```
jgpruitt, line 264: > This flag means that everything will be owned by the tsdbadmin user in the target

IIF tsdbadmin is used to run pg_restore

use-timescale/migration/dual-write-and-backfill.md
jgpruitt: yeah, that sounds like a good idea

jgpruitt (changes requested):

use-timescale/migration/dual-write-and-backfill.md
alejandrodnm: WDYT about not mentioning tsdbadmin? Maybe something like:

> b. `--no-owner` is required because assigning ownership to certain objects
   could necessitate superuser permissions. Such permissions are not supported
   by Timescale services. When using this flag, the ownership of migrated
   objects will automatically be assigned to the role that is used to run the
   command, regardless of the original ownership in the source database. This is
   a limitation.

JamesGuthrie: @alejandrodnm I think that suggestion is good. I've included something like it in my revision.

use-timescale/migration/dual-write-and-backfill.md
JamesGuthrie: I think that the suggestion is generally correct, but I would suggest a very different approach, which includes something which is as yet missing in this text. I will push a revision with those suggestions.

JamesGuthrie: To clarify: IMHO we should suggest that people who are coming from plain postgres and who have large time-series tables convert those tables to hypertables. This means that they will exclude the data from those tables in their pg_dump, and backfill the tables into the hypertable.

PR #2709, opened 2023-09-26
Adjust schema for start_background_workers to new location

svenklemm:


mkindahl (approved):

antekresic (approved):
There are a few other references to `_timescaledb_internal` functions in the docs. If you could add those as well, would be great.

PR #2751, opened 2023-10-20
Add clarification about read-replica connection during failover

mfreed:
# Description

Clarify that the read-replica connection is not highly available for reads during replica recovery.


ggodeke (approved):
LGTM

PR #2752, opened 2023-10-21
Clarify that ALTER compression command does not compress itself

mfreed:
# Description

- Clarify that ALTER compression command should often be used  in combination with compression policy or compress_chunk
- Clarify that segmentby is important field

### DIFF
--- a/api/alter_table_compression.md
+++ b/api/alter_table_compression.md
@@ -16,7 +16,7 @@ options.
 
 By itself, this `ALTER` statement alone does not compress a hypertable. To do so, either create a
 compression policy using the [add_compression_policy][add_compression_policy] function or manually
-compress a specific hypertable chunk via the [compress_chunk][compress_chunk] function.
+compress a specific hypertable chunk using the [compress_chunk][compress_chunk] function.
 
 The syntax is:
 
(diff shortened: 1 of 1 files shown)


api/alter_table_compression.md
```diff
@@ -46,7 +52,9 @@ timescaledb.compress_segmentby = '<column_name> [, ...]', timescaledb.compress_c
 
 ## Sample usage
 
-Configure a hypertable that ingests device data to use compression:
+Configure a hypertable that ingests device data to use compression. Here, because we will
+often query the hypertable about a specific device or set of devices, we segment the
+compression using the `device_id` for greater performance.
```
horzsolt, line 57: Maybe we can include a link to the segmentby description.

```suggestion
compression using the `device_id` for [greater performance](https://docs.timescale.com/use-timescale/latest/compression/about-compression/#segment-by-columns).
```

horzsolt (approved):
added a link proposal about the segmentby, the rest looks good.

api/alter_table_compression.md
mfreed: That link doesn't work for me.

It redirects here, which doesn't have anything:

https://docs.timescale.com/use-timescale/latest/compression/#segment-by-columns

PR #2755, opened 2023-10-23
Expectations on the maintenance behavior with a replica.

alexeyklyukin:
# Description

Correct the description of the maintenance behavior when replicas are present. In the previous version, we promised our customers no downtime and two failover events, while in reality, the maintenance may result in up to two write downtimes, up to one minute each, since the replica waits for up to a minute for the primary to restart.

# Links

See https://iobeam.slack.com/archives/CDZGF8U7J/p1697611316402379

# Review checklists

### DIFF
--- a/use-timescale/upgrades.md
+++ b/use-timescale/upgrades.md
@@ -48,17 +48,23 @@ applied when your service is resumed.
 
 ## Replicas and maintenance
 
-Services with replicas do not require maintenance downtime. Instead, they have
-one or two failover events during maintenance, taking less than a few seconds
-each.
+Services with replicas require minimal write downtime during maintenance,
+while read-only queries keep working through the maintenance. The maintenance
+requires up to two failovers, performed automatically, taking less than a few
+seconds each.
 
 During a maintenance event, services with replicas perform maintenance on each
-node independently. Maintenance begins on one node, and when it is finished,
-that node is promoted to primary. The other node then begins maintenance, and
-when it is complete, it remains the replica. Sometimes, maintenance begins with
-the primary. This causes the replica node to be promoted at the start. If this
-happens, the service experiences two promotions, or failovers, during the
-maintenance event.
+node independently. When maintenance goes on with the primary node, the primary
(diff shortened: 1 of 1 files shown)


use-timescale/upgrades.md
```diff
@@ -48,17 +48,23 @@ applied when your service is resumed.
...
-that node is promoted to primary. The other node then begins maintenance, and
-when it is complete, it remains the replica. Sometimes, maintenance begins with
-the primary. This causes the replica node to be promoted at the start. If this
-happens, the service experiences two promotions, or failovers, during the
-maintenance event.
+node independently. When maintenance goes on with the primary node, the primary
+node needs to be restarted. If the restart takes more than a minute, the replica
+node is promoted to the primary, given that the replica has no replication lag.
+If the maintenance on the primary node is completed within a minute and it comes
+back online, the replica remains as a replica.
```
inchuev, line 61: Maybe we should mention the reason behind this - namely, to prevent connections from being torn down twice.

use-timescale/upgrades.md
```diff
@@ -48,17 +48,23 @@ applied when your service is resumed.
...
+
+When maintenance starts on the primary and results in the promotion of the replica
+node, the maintenance proceeds then with the newly promoted replica, following the same
+sequence. If the newly promoted replica takes more than a minute to restart, the former
```
inchuev, line 65: And if there's no replication lag, I presume?

inchuev (approved):
Thank you for the corrections!

use-timescale/upgrades.md
alexeyklyukin: I think the reason is more in "how" rather than "why". We would prefer to manually failover just before bouncing pods, but we can't do this with the `rollingUpdate` strategy. In the worst case, we would still have 2 connection bounces with this strategy, the same as if we were doing switchovers pre-emptively.

use-timescale/upgrades.md
alexeyklyukin: Yeah, I decided not to duplicate it here, since we mention "the same sequence".

PR #2759, opened 2023-10-24
Add live migration from postgres

alejandrodnm:
# Description

# Links

# Writing help

# Review checklists


migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+Before beginning the migration process, ensure that the following tools are
+installed and available on the system that performs the migration:
+
+- `psql`, `pg_dump`, and `pg_dumpall`: These can be installed by following the
+  instructions in the [How to Install psql on Mac, Ubuntu, Debian, Windows][install-psql]
+  blog post. Although the instructions specifically mention `psql`, following
+  them also installs `pg_dump` and `pg_dumpall`.
+
+- `pgcopydb`: Installation instructions can be found in the
+  [official repository][install-pgcopydb].
```
JamesGuthrie, line 28: I would prefer it if we don't send people away from our docs page for them to install our tools. Could we perhaps provide installation instructions for the one/two most common distros, and provide links for people using other distros?

It looks like we should be able to provide one set of instructions for ubuntu and debian, so I would suggest that we do that. I briefly tried to install postgres on rocky linux, and it failed with a gpg signature error, so I don't think that's worth investigating.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+1. Promote target database as new primary.
+
+<GettingHelp />
+
+<StepOne />
+
+## 2. Set up a replication slot and snapshot
+
+Use the `pgcopydb follow` to create a [replication slot][replication-slot] and
+a [snapshot][snapshot].
```
JamesGuthrie, line 52: I would leave the links off of these terms here. We provide more context below, and have a link there. The link here tempts me to click it and break out of the flow. With the additional context below, I can make a better-informed decision about whether I need to leave the current page.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+
+```sh
+pgcopydb follow \
+  --source "$SOURCE" \
+  --target "$TARGET" \
+  --fail-fast \
+  --plugin wal2json
```
JamesGuthrie, line 59: (Questios for my own interest): Is wal2json available in all Postgres versions? What is the minimum Postgres version that we need for live migration?

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+1. Apply replication changes.
+1. Promote target database as new primary.
+
+<GettingHelp />
+
+<StepOne />
+
+## 2. Set up a replication slot and snapshot
+
+Use the `pgcopydb follow` to create a [replication slot][replication-slot] and
```
JamesGuthrie, line 51: It's unclear where these commands should be executed, or what the requirements of that machine are (available storage space etc.).

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+strategy.
+
+> A slot represents a stream of changes that can be replayed to a client in the
+order they were made on the origin server.
+
+The stream of changes emitted by the slot are buffered into disk until they are
+applied on the target. The instance used to orchestrate the migration (the one
+running the commands) should have enough capacity to store the files, and it
+should be actively monitored to prevent any issues that might result due to
+lack of space.
```
JamesGuthrie, line 81: Ah, I mentioned that this is missing in a comment above. I think that this content needs to be moved further up within this section.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+
+<!-- TODO:  mention the housekeeping script that clears the buffered file s -->
+
+Before the stream of changes can be applied, the schema and data that existed
+prior to the creation of the replication slot must be migrated ([step 3][step-3]). The
+point that marks the beginning of the replication and buffering of changes is
+given by the exported snapshot. The larger the database, the more time it takes
+to perform the initial migration, and the longer the buffered files need to be
+stored.
```
JamesGuthrie, line 90: I think that we would be well-served by adding a diagram somewhere to explain the overall process. We're working on one for the blog post, which I think we can then put here as well.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+
+## 3. Migrate roles and schema from source to target
+
+### 3a. Dump the database roles from the source database
+
+<DumpDatabaseRoles />
```
JamesGuthrie, line 96: This doesn't pull anything in.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+```sh
+pg_dump -d "$SOURCE" \
+  --format=plain \
+  --quote-all-identifiers \
+  --no-tablespaces \
+  --no-owner \
+  --no-privileges \
+  --schema-only \
+  --file=dump.sql \
+  --snapshot=$(cat /tmp/pgcopydb/snapshot)
```
JamesGuthrie, line 109: Above you mention that this path is driven by $TMPDIR
```suggestion
  --snapshot=$(cat ${TMPDIR}/pgcopydb/snapshot)
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+- `--schema-only` is used to dump only the object definitions (schema), not
+  data.
+
+- `--snapshot` is use to specified the synchronized [snapshot][snapshot] when
+  making a dump of the database.
+
+<ExplainPgDumpFlags />
```
JamesGuthrie, line 118: This doesn't pull anything in.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+  --echo-errors \
+  -f roles.sql \
+  -f dump.sql
+```
+
+## 4. Convert hypertables and enable Timescale features
+
+This is the ideal point to convert regular tables into hypertables. This step
+is optional. It can even be performed after all the data has been migrated,
+since hypertables can be created from tables with existing data.
```
JamesGuthrie, line 134: I would be cautious with the statement. It is correct, but it misses the fact that if you convert a plain table to a hypertable after-the-fact, the table can't accept writes through the duration of the conversion process, which is basically the equivalent to having application downtime.

JamesGuthrie: I see that you have a bit more detailed description below. I would rework this section to say something along the lines of "while it is optional to set up hypertables before moving data, if you do it later it's going to be a pain in the ass and you're going to regret it".

migrate/live-migration/live-migration-from-postgres.md
JamesGuthrie: Coming back to this, I realized that if $TMPDIR is not set then it's obviously incorrect to do this 🙈  In that case, I would suggest that we either skip the whole $TMPDIR explanation and never mention it, or that we _set_ $TMPDIR and consistently use it throughout. My preference is for the former.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+Use `pgcopydb` and the snapshot ID obtained in the second step to copy the
+data. The [snapshot][snapshot] ID is exported to `${TMPDIR}/pgcopydb/snapshot`.
+If `$TMPDIR` is not set, it's exported to `/tmp/pgcopydb/snapshot`.
+
+```sh
+pgcopydb copy table-data \
+  --source "$SOURCE" \
+  --target "$TARGET" \
+  --snapshot $(cat /tmp/pgcopydb/snapshot) \
+  --split-tables-larger-than <same-table concurrency size threshold>  \
```
JamesGuthrie, line 169: How should this be configured?

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+WHERE application_name='pgcopydb' AND state='streaming'
+EOF
+```
+
+- `replay_lag`: Time elapsed between flushing recent WAL locally and receiving
+  notification that this standby server has written, flushed and applied it.
+
+Before proceeding to the next step, you need to wait until the replication lag
+is minimized as much as possible. For instance, you can move forward when the
+`replay_lag` is close to just a few seconds.
```
JamesGuthrie, line 199: What if the replay lag never gets close to just a few seconds?

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,252 @@
...
+## 7. Promote target database as new primary
+
+At this point, the source and target should be almost identical, with the
+`replay_lag` close to just a few seconds.
+
+When it's time to perform the switchover, stop the write traffic to the source
+database.
+
+Since no new data is being ingested, `pgcopydb` can be instructed to stop the
+replication process.
```
JamesGuthrie, line 210: I think that we need to be cautious here. If the replication lag _doesn't_ go down to a few seconds, then the user _must_ wait for it to go down here. We don't mention this, and it could lead to data loss, which we don't want.

JamesGuthrie (changes requested):

migrate/live-migration/live-migration-from-postgres.md
JamesGuthrie: Also: apparently pgcopydb isn't available for RHEL-based distros, which backs the idea of providing one set of instructions for debian/ubuntu and leaving the RHEL-strugglers to struggle.

migrate/live-migration/live-migration-from-postgres.md
alejandrodnm: I just did a quick read and it's an external extension, I think it's installed with the postgres-dev package. From the pgcopydb's docs I get that we can use `test_decoding` and that one I think comes bundle with postgres and it's the default for pgcopydb.

@arajkumar is there a reason why we are using wal2json and not the other? WDYT

migrate/live-migration/live-migration-from-postgres.md
alejandrodnm: @arajkumar can you check up on this?

migrate/live-migration/live-migration-from-postgres.md
alejandrodnm: I'd say you either run out of space, or remain on a state of constant lag, or you have to stop ingest to burn down the lag.

alejandrodnm: I'll move this part to the next section so we can tackle all the lag nuances there.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,321 @@
...
+
+For Debian and Ubuntu systems, you can install all the tools with:
+
+```
+sudo apt update
+sudo apt install -y postgresql-common
+sudo /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh
+sudo apt install -y pgcopydb
+```
+
```
JamesGuthrie, line 33:
```suggestion

On other distributions, you can use the following installation instructions:
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,321 @@
...
+> A slot represents a stream of changes that can be replayed to a client in the
+order they were made on the origin server.
+
+The stream of changes emitted by the slot are buffered into disk until they are
+applied on the target. The instance used to orchestrate the migration (the one
+running the commands) should have enough capacity to store the files, and it
+should be actively monitored to prevent any issues that might result due to
+lack of space.
+
+Use the `pgcopydb`'s follow command to create a replication slot:
```
JamesGuthrie, line 77:
```suggestion
Use `pgcopydb`'s follow command to create a replication slot:
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,321 @@
...
+  --plugin wal2json > pgcopydb_follow.log 2>&1 &
+```
+
+The `> pgcopydb_follow.log 2>&1` part redirects all the messages to the
+`pgcopydb_follow.log` file, this is optional but recommended. The `pgcopydb
+follow` command outputs many messages, if they are not redirected, using the
+terminal becomes cumbersome due to the constant pop up of messages.
+
+The `follow` command not only creates the replication slot for streaming
+changes, but also exports a [snapshot][snapshot] ID to `tmp/pgcopydb/snapshot`.
```
JamesGuthrie, line 108:
```suggestion
changes, but also exports a [snapshot][snapshot] ID to `/tmp/pgcopydb/snapshot`.
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,321 @@
...
+  --no-privileges \
+  --schema-only \
+  --file=dump.sql \
+  --snapshot=$(cat /tmp/pgcopydb/snapshot)
+```
+
+- `--schema-only` is used to dump only the object definitions (schema), not
+  data.
+
+- `--snapshot` is use to specified the synchronized [snapshot][snapshot] when
```
JamesGuthrie, line 148:
```suggestion
- `--snapshot` is used to specified the synchronized [snapshot][snapshot] when
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,321 @@
...
+The disadvantage of converting after inserting all the data is that the table
+must be locked while the data is being partitioned. This means that no new data
+can be written into the table while the table is being converted. The larger
+the table, the longer it is going to remain locked. Conversely, inserting into
+an already converted Hypertable is a fast operation that partitions the data
+while it's being inserted.
+
+## 5. Migrate initial data from source to target
+
+Use `pgcopydb` and the snapshot ID obtained in the [step 2][step-2]
```
JamesGuthrie, line 196:
```suggestion
Use `pgcopydb` and the snapshot ID obtained in [step 2][step-2]
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,321 @@
...
+
+```sh
+pgcopydb copy table-data \
+  --source "$SOURCE" \
+  --target "$TARGET" \
+  --snapshot $(cat /tmp/pgcopydb/snapshot) \
+  --split-tables-larger-than <same-table concurrency size threshold>  \
+  --table-jobs <jobs: usually the number of core in SOURCE>
+```
+
```
JamesGuthrie, line 207: This command runs for (presumably) a long time. Should we give users a warning about that? And tell them to wait for this command to complete before proceeding to the next step?

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,321 @@
...
+```sh
+psql $SOURCE \
+  -f - <<'EOF'
+SELECT replay_lag
+FROM pg_stat_replication
+WHERE application_name='pgcopydb' AND state='streaming'
+EOF
+```
+
+- `replay_lag`: Time elapsed between flushing recent WAL locally and receiving
```
JamesGuthrie, line 229: What is the unit? Seconds? Milliseconds?

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,321 @@
...
+  notification that this standby server has written, flushed and applied it.
+
+## 7. Promote target database as new primary
```
JamesGuthrie, line 232: In general I think that we're missing giving the user enough context about how the whole migration process works. IMO that would be the function of the parent page (Low-downtime: Live migration), which is currently absent.

Aside from the missing context, I think this section has all of the information that it needs, but I think that it needs to be reworked a little for clarity. I think that it would only make sense to do so once we've got the content for the context. Maybe some things will become clearer then.

JamesGuthrie (commented):
Getting there! I think the main feedback is the missing context/parent page which explains what Live migration actually is.

migrate/live-migration/live-migration-from-postgres.md
arajkumar: There was a bug in pgcopydb while using with test_decoding which prevents migration of a table which has reserved Postgres keywords as column name.(e.g. time). I [fixed it in upstream](https://github.com/dimitri/pgcopydb/pull/482), but it is not yet released to the public.

We have to stick to wal2json until  the new pgcopydb release is made available.

wal2json is usually distributed with most of the postgres distributions. Most of the cloud providers(AWS, GCP, Azure) has enabled wal2json in their managed offerings.

migrate/live-migration/live-migration-from-postgres.md
arajkumar: --split-tables-larger-than

Allow [Same-table Concurrency](https://pgcopydb.readthedocs.io/en/latest/design.html#same-table-concurrency) when processing the source database. This environment variable value is expected to be a byte size, and bytes units B, kB, MB, GB, TB, PB, and EB are known.


Reference: https://pgcopydb.readthedocs.io/en/latest/ref/pgcopydb_clone.html

migrate/live-migration/live-migration-from-postgres.md
alejandrodnm: The type is interval so it's based on the intervalstyle variable, by default it's pretty printed and depending on how big it's there's more or less information:

https://www.postgresql.org/docs/16/datatype-datetime.html#DATATYPE-INTERVAL-OUTPUT

```
adn@/tmp:adn> select interval '1 second'
╒══════════╕
│ interval │
╞══════════╡
│ 0:00:01  │
╘══════════╛
SELECT 1
Time: 0.004s
adn@/tmp:adn> select interval '1 month'
╒══════════════════╕
│ interval         │
╞══════════════════╡
│ 30 days, 0:00:00 │
╘══════════════════╛
```

migrate/live-migration/live-migration-from-postgres.md
alejandrodnm: I think we should set this to a default like 10GB and add a note about it. WDYT?

alejandrodnm: And table-jobs to 8?

JamesGuthrie: Yup, that sounds good

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,370 @@
...
+
+<SourceTargetNote />
+
+In detail, the migration process consists of the following steps:
+
+1. Set up a target database instance in Timescale.
+1. Prepare the source database for the live migration.
+1. Set up a replication slot and snapshot.
+1. Migrate roles and schema from source to target.
+1. Convert hypertables and enable Timescale features.
```
VineethReddy02, line 54: What do we mean by "enable Timescale features"?

Lets changes this to "Enable hypertables" instead of "Convert Hypertables" as this sounds like convert existing hyertables to something else.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,370 @@
...
+1. Migrate initial data from source to target.
+1. Apply replication changes.
```
VineethReddy02, line 56: Maybe change this to "Apply change data capture (CDC) on target"

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,370 @@
...
+strategy.
+
+> A slot represents a stream of changes that can be replayed to a client in the
+order they were made on the origin server.
+
+The stream of changes emitted by the slot are buffered into disk until they are
+applied on the target. The instance used to orchestrate the migration (the one
+running the commands) should have enough capacity to store the files, and it
+should be actively monitored to prevent any issues that might result due to
+lack of space.
```
VineethReddy02, line 113: Should we recommend size of ec2 instance? that is compute and memory and disk that they should provision? For disk maybe a simple calculation based on real time inserts/updates/deletes.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,370 @@
...
+  -v ON_ERROR_STOP=1 \
+  -c "SELECT create_hypertable('<table name>', '<time column name>')"
+```
+
+A more detailed explanation can be found in the [hypertable
+documentation][hypertable-docs].
+
+Once the table is converted, you can follow the guides to enable more Timescale
+features like [retention][retention-docs], [compression][compression-docs], and
+[data tiering][data-tiering-docs].
```
VineethReddy02, line 219: I don't we should refer to data tiering at this point. We need to test this before we mention here. 

I want to avoid confusion/segways for the users following this doc.

migrate/live-migration/live-migration-from-postgres.md
VineethReddy02: cc @arajkumar

VineethReddy02 (changes requested):
Also, we have to update `/migrate` (https://docs.timescale.com/migrate/latest/) parent page with "Live migration" availability alongside dump. restore and dual writes and backfill.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,370 @@
...
+If the query returns something other than `logical`, run the following command
+to set it up:
+
+```sh
+psql -X -d $SOURCE -c 'alter system set wal_level=logical'
+```
+
+You should restart your database for the changes to take effect.
+
+## 3. Set up a replication slot and snapshot
```
arajkumar, line 101: We need to mention about [REPLICA IDENTITY](https://www.postgresql.org/docs/16/sql-altertable.html#SQL-ALTERTABLE-REPLICA-IDENTITY) to support replication of DELETE & UPDATE operations.

migrate/live-migration/live-migration-from-postgres.md
arajkumar: Ideally, 4 CPU & 4 to 8 GiB of RAM with 1.2 TiB of disk (for 10K/s ingestion load with the assumption of historical data copy takes 2 days).

arajkumar (approved):
lgtm, if possible adding a note about REPLICATION IDENTITY would be great!

migrate/live-migration/live-migration-from-postgres.md
alejandrodnm: By features I meant retention, compression ... I'll change it `Enable hypertables` sounds better 👍

migrate/live-migration/live-migration-from-postgres.md
alejandrodnm: We never mention CDC or what does it mean. If you want to put it there as a title we'd have to explain what it is. Maybe we can change it to `Apply stream of changes coming from source`

VineethReddy02: "stream of changes", SGTM!

migrate/index.md
```diff
@@ -35,8 +35,9 @@ migrating from PostgreSQL.
 
 If you're using PostgreSQL, you may also have heard of logical replication
 being the recommended strategy for migrations with low downtime. Currently,
-TimescaleDB doesn't work with logical replication, so this is not a viable
-option, but we are actively working on making this possible.
+TimescaleDB supports logical replication only from sources that are using
```
VineethReddy02, line 38: This isn't true but this [PR](https://github.com/timescale/docs/pull/2763) should address this.

arajkumar (approved):
lgtm

JamesGuthrie (approved):
I've addressed most of the comments myself in my most recent commit.

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,399 @@
...
+
+```sh
+psql -X -d $SOURCE -c 'show old_snapshot_threshold'
+```
+
+If the query returns something other than `-1`, run the following command
+to set it up:
+
+```sh
+psql -X -d $SOURCE -c 'alter system set old_snapshot_threshold=-1'
```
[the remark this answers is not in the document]
JamesGuthrie: done

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+If the query returns something other than `logical`, run the following command
+to set it up:
+
+```sh
+psql -X -d $SOURCE -c 'alter system set wal_level=logical'
+```
+
+You should restart your database for the changes to take effect.
```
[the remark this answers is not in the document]
JamesGuthrie: done

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -0,0 +1,399 @@
...
+
+This document provides detailed step-by-step instructions to migrate data using
+[pgcopydb][pgcopydb] to perform a live migration from a source database which
+is using PostgreSQL to Timescale.
+
+It's suggested to provision a dedicated instance to run the migration steps.
+Ideally an AWS EC2 instance that's in the same region as the Timescale target
+service. For a 10K transactions/s ingestion load, with the assumption of
+historical data copy takes 2 days, the recommend is 4 CPU with 4 to 8 GiB of
+RAM and a 1.2 TiB of storage.
```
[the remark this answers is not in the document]
JamesGuthrie: done

VineethReddy02 (approved):

PR #2761, opened 2023-10-27
Fix documentation about max_locks_per_transaction

svenklemm:
max_locks_per_transaction is multiplied by max_connections which is the hard limit for how many locks can exist at the same time.


jkoria (approved):
Looks good.

PR #2820, opened 2023-11-16
Adjust TimescaleDB version match for migration

JamesGuthrie:
# Description

Timescale doesn't support downgrading TimescaleDB or Postgres versions, so users must upgrade the source DB before migrating.


alejandrodnm:
Should we point users to https://docs.timescale.com/self-hosted/latest/upgrades/ ?

alejandrodnm (approved):

PR #2897, opened 2023-12-18
feat: document live migration from TimescaleDB to Timescale

harkishen:
# Description

This PR adds live migration documentation for TimescaleDB to Timescale migration and modifies the existing Postgres to Timescale migration documentation by using live migration docker image. The live migration image simplifies the migration process by orchestrating steps involved in CDC process, providing a seamless migration experience.

# Links

Fixes https://github.com/timescale/team-data-onboarding/issues/40

# Review checklists

### DIFF
--- a/_partials/_migrate_live_migration_step2.md
+++ b/_partials/_migrate_live_migration_step2.md
@@ -1,3 +1,7 @@
+import SetupSourceTarget from "versionContent/_partials/_migrate_set_up_source_and_target.mdx";
+
+<SetupSourceTarget />
+
 It's important to ensure that the `old_snapshot_threshold` value is set to the
 default value of `-1` in your source database. This prevents PostgreSQL from
 treating the data in a snapshot as outdated. If this value is set other than
--- a/_partials/_migrate_live_migration_step2.md
+++ b/_partials/_migrate_live_migration_step2.md
@@ -1,0 +1,44 @@
+It's important to ensure that the `old_snapshot_threshold` value is set to the
+default value of `-1` in your source database. This prevents PostgreSQL from
+treating the data in a snapshot as outdated. If this value is set other than
+`-1`, it might affect the existing data migration step.
+
+To check the current value of `old_snapshot_threshold`, run the command:
+
+```sh
+psql -X -d $SOURCE -c 'show old_snapshot_threshold'
+```
+
+If the query returns something other than `-1`, you must change it.
+
+If you have a superuser on a self-hosted database, run the following command:
+
+```sh
+psql -X -d $SOURCE -c 'alter system set old_snapshot_threshold=-1'
+```
+
+Otherwise, if you are using a managed service, use your cloud provider's
+configuration mechanism to set `old_snapshot_threshold` to `-1`.
+
+Next, you should set `wal_level` to `logical` so that the write-ahead log (WAL)
+records information that is needed for logical decoding.
+
+To check the current value of  `wal_level`, run the command:
+
+```sh
+psql -X -d $SOURCE -c 'show wal_level'
+```
+
+If the query returns something other than `logical`, you must change it.
+
+If you have a superuser on a self-hosted database, run the following command:
+
+```sh
+psql -X -d $SOURCE -c 'alter system set wal_level=logical'
+```
+
+Otherwise, if you are using a managed service, use your cloud provider's
+configuration mechanism to set `wal_level` to `logical`.
+
+Restart your database for the changes to take effect, and verify that the
+settings are reflected in your database.
--- a/migrate/live-migration/index.md
+++ b/migrate/live-migration/index.md
@@ -10,10 +10,9 @@ import SourceTargetNote from "versionContent/_partials/_migrate_source_target_no
 # Live migration
 
 Live migration is a migration strategy to move a large amount of data
-(100&nbsp;GB-10&nbsp;TB+) with low downtime (on the order of minutes of
-downtime). It is significantly more complicated to execute than a migration
-with downtime using [pg_dump/restore][pg-dump-and-restore], but supports more
-use-cases and has less requirements than the [dual-write and backfill] method.
+(100&nbsp;GB-10&nbsp;TB+) with low downtime (in the order of few minutes). It requires
+more steps to execute than a migration with downtime using [pg_dump/restore][pg-dump-and-restore],
+but supports more use-cases and has less requirements than the [dual-write and backfill] method.
 
 <SourceTargetNote />
 
@@ -27,20 +26,23 @@ Roughly, it consists of four steps:
 3. Copy data from source to target while capturing changes.
 4. Apply captured changes from source to target.
 
-Currently live migration only supports migrating from PostgreSQL, but we are
-actively working on supporting TimescaleDB.
-
 Live migration works well when:
 - Large, busy tables have primary keys, or don't have many `UPDATE` or
   `DELETE` statements.
-- The insert workload does not exceed 20'000 rows per second, and
+- The insert workload does not exceed 20,000 rows per second, and
   inserts are batched. If your application exceeds this, you should use
   the [dual-write and backfill] migration method.
 
-For more information, consult the step-by-step migration guide:
+For more information, refer to the step-by-step migration guide:
 
 - [Live migration from PostgreSQL][from-postgres]
+- [Live migration from TimescaleDB][from-timescaledb]
+
+If you are migrating from AWS RDS to Timescale, you can refer to [this][live-migration-playbook] playbook
+for a detailed migration guide.
 
 [from-postgres]: /migrate/:currentVersion:/live-migration/live-migration-from-postgres/
+[from-timescaledb]: /migrate/:currentVersion:/live-migration/live-migration-from-timescaledb/
 [pg-dump-and-restore]: /migrate/:currentVersion:/pg-dump-and-restore/
 [dual-write and backfill]: /migrate/:currentVersion:/dual-write-and-backfill/
+[live-migration-playbook]: /migrate/:currentVersion:/playbooks/rds-timescale-live-migration/
--- a/migrate/live-migration/index.md
+++ b/migrate/live-migration/index.md
@@ -10,7 +10,7 @@ import SourceTargetNote from "versionContent/_partials/_migrate_source_target_no
 # Live migration
 
 Live migration is a migration strategy to move a large amount of data
-(100&nbsp;GB-10&nbsp;TB+) with low downtime (in the order of few minutes). It requires
+(100&nbsp;GB-10&nbsp;TB+) with low downtime (on the order of few minutes). It requires
 more steps to execute than a migration with downtime using [pg_dump/restore][pg-dump-and-restore],
 but supports more use-cases and has less requirements than the [dual-write and backfill] method.
 
@@ -38,11 +38,15 @@ For more information, refer to the step-by-step migration guide:
 - [Live migration from PostgreSQL][from-postgres]
 - [Live migration from TimescaleDB][from-timescaledb]
 
+If you want to manually migrate data from PostgreSQL, refer to
+[Live migration from PostgreSQL manually][live-migration-manual].
+
 If you are migrating from AWS RDS to Timescale, you can refer to [this][live-migration-playbook] playbook
 for a detailed migration guide.
 
 [from-postgres]: /migrate/:currentVersion:/live-migration/live-migration-from-postgres/
 [from-timescaledb]: /migrate/:currentVersion:/live-migration/live-migration-from-timescaledb/
+[live-migration-manual]: /migrate/:currentVersion:/playbooks/live-migration-from-timescaledb-manually/
 [pg-dump-and-restore]: /migrate/:currentVersion:/pg-dump-and-restore/
 [dual-write and backfill]: /migrate/:currentVersion:/dual-write-and-backfill/
 [live-migration-playbook]: /migrate/:currentVersion:/playbooks/rds-timescale-live-migration/
--- a/migrate/live-migration/live-migration-from-postgres.md
+++ b/migrate/live-migration/live-migration-from-postgres.md
@@ -12,6 +12,7 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
 import DumpDatabaseRoles from "versionContent/_partials/_migrate_dual_write_dump_database_roles.mdx";
 import DumpPreDataSourceSchema from "versionContent/_partials/_migrate_pre_data_dump_source_schema.mdx";
 import DumpPostDataSourceSchema from "versionContent/_partials/_migrate_post_data_dump_source_schema.mdx";
+import LiveMigrationStep2 from "versionContent/_partials/_migrate_live_migration_step2.mdx";
 
 # Live migration from PostgreSQL database with pgcopydb
 
@@ -69,50 +70,7 @@ In detail, the migration process consists of the following steps:
 
 ## 2. Prepare the source database for the live migration
 
-It's important to ensure that the `old_snapshot_threshold` value is set to the
-default value of `-1` in your source database. This prevents PostgreSQL from
-treating the data in a snapshot as outdated. If this value is set other than
-`-1`, it might affect the existing data migration step.
-
-To check the current value of `old_snapshot_threshold`, run the command:
-
-```sh
-psql -X -d $SOURCE -c 'show old_snapshot_threshold'
-```
-
-If the query returns something other than `-1`, you must change it.
-
-If you have a superuser on a self-hosted database, run the following command:
-
-```sh
-psql -X -d $SOURCE -c 'alter system set old_snapshot_threshold=-1'
-```
-
-Otherwise, if you are using a managed service, use your cloud provider's
-configuration mechanism to set `old_snapshot_threshold` to `-1`.
-
-Next, you should set `wal_level` to `logical` so that the write-ahead log (WAL)
-records information that is needed for logical decoding.
-
-To check the current value of  `wal_level`, run the command:
-
-```sh
-psql -X -d $SOURCE -c 'show wal_level'
-```
-
-If the query returns something other than `logical`, you must change it.
-
-If you have a superuser on a self-hosted database, run the following command:
-
-```sh
-psql -X -d $SOURCE -c 'alter system set wal_level=logical'
-```
-
-Otherwise, if you are using a managed service, use your cloud provider's
-configuration mechanism to set `wal_level` to `logical`.
-
-Restart your database for the changes to take effect, and verify that the
-settings are reflected in your database.
+<LiveMigrationStep2 />
 
 ## 3. Set up a replication slot and snapshot
 
--- a/migrate/live-migration/live-migration-from-postgres.md
+++ b/migrate/live-migration/live-migration-from-postgres.md
@@ -16,356 +16,153 @@ import LiveMigrationStep2 from "versionContent/_partials/_migrate_live_migration
 
 # Live migration from PostgreSQL database with pgcopydb
 
-This document provides detailed step-by-step instructions to migrate data using
-[pgcopydb][pgcopydb] to perform a live migration from a source database which
-is using PostgreSQL to Timescale.
+This document provides detailed instructions to migrate data from your
+PostgreSQL database to a Timescale instance with minimal downtime (on the order
+of a few minutes) of your production applications, using the [live migration]
+strategy. To simplify the migration, we provide you with a docker image
+containing all the tools and scripts that you need to perform the live
+migration.
 
 You should provision a dedicated instance to run the migration steps from.
 Ideally an AWS EC2 instance that's in the same region as the Timescale target
-service. For an ingestion load of 10'000 transactions/s, and assuming that the
+service. For an ingestion load of 10,000 transactions/s, and assuming that the
 historical data copy takes 2 days, we recommend 4 CPUs with 4 to 8 GiB of RAM
 and 1.2 TiB of storage.
-
-Before beginning the migration process, ensure that tools `psql`, `pg_dump`,
-`pg_dumpall`, and `pgcopydb` are installed and available on the system that
-performs the migration.
-
-For Debian and Ubuntu systems, you can install all the tools with:
-
-```
-sudo apt update
-sudo apt install -y postgresql-common
-sudo /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh
-sudo apt install -y pgcopydb
-```
-
-On other distributions, you can use the following installation instructions:
-
-- `pgcopydb`: Installation instructions can be found in the [official
-  repository][install-pgcopydb]. When installing from package managers like
-  `apt`, `yum`, or `dnf`, the other required tools are usually also installed
-  as dependencies to `pgcopydb`.
-
-- `psql`, `pg_dump`, and `pg_dumpall`: These can be installed by following the
-  instructions in the [How to Install psql on Mac, Ubuntu, Debian, Windows][install-psql]
-  blog post. Although the instructions specifically mention `psql`, following
-  them also installs `pg_dump` and `pg_dumpall`.
 
 <SourceTargetNote />
 
 In detail, the migration process consists of the following steps:
 
 1. Set up a target database instance in Timescale.
-1. Prepare the source database for the live migration.
-1. Set up a replication slot and snapshot.
-1. Migrate roles and schema from source to target.
-1. Enable hypertables.
-1. Migrate initial data from source to target.
-1. Apply the replicated changes from source.
-1. Promote target database as new primary.
+1. Prepare the source database for live migration.
+1. Run the live migration docker image.
+1. Validate the data in target database and use it as new primary.
 
 <GettingHelp />
 
 <StepOne />
 
-## 2. Prepare the source database for the live migration
+## 2. Prepare the source database for live migration
 
 <LiveMigrationStep2 />
 
-## 3. Set up a replication slot and snapshot
+Next, you need to ensure that your source tables and hypertables have either a primary key
+or `REPLICA IDENTITY` set. This is important as it is a requirement for replicating `DELETE` and
+`UPDATE` operations. Replica identity assists the replication process in identifying the rows
+being modified. It defaults to using the table's primary key.
 
-The [replication slot][replication-slot] forms the backbone of the replication
-strategy.
+If a table doesn't have a primary key, you'll have to manually set the replica identity.
+One option is to use a unique, non-partial, non-deferrable index that includes only columns
+marked as `NOT NULL`. This can be set as the replica identity:
 
-> A slot represents a stream of changes that can be replayed to a client in the
-order they were made on the origin server.
-
-The stream of changes emitted by the slot are buffered into disk until they are
-applied on the target. The instance used to orchestrate the migration (the one
-running the commands) should have enough capacity to store the files, and it
-should be actively monitored to prevent any issues that might result due to
-lack of space.
-
-Before starting, there's an important caveat. To replicate `DELETE` and
-`UPDATE` operations, the source table must either have a primary key or
-`REPLICA IDENTITY` set. Replica identity assists logical decoding in
-identifying the rows being modified. It defaults to using the table's primary
-key.
-
-If a table doesn't have a primary key, you'll have to manually set the replica
-identity. One option is to use a unique, non-partial, non-deferrable index that
-includes only columns marked as `NOT NULL`. This can be set as the replica
-identity:
-
-```sql
+```sh
 ALTER TABLE {table_name} REPLICA IDENTITY USING INDEX {_index_name}
 ```
 
-If there's no primary key or viable unique index to use, you'll have to set
-`REPLICA IDENTITY` to `FULL`. If you are expecting a large number of `UPDATE`
-or `DELETE` operations on the table we **do not recommend** using `FULL`. For
-each `UPDATE` or `DELETE` statement, Postgres will have to read the whole table
-to find all matching rows, which will result in significantly slower
-replication.
+If there's no primary key or viable unique index to use, you have to set `REPLICA IDENTITY`
+to `FULL`. If you are expecting a large number of `UPDATE` or `DELETE` operations on the table,
+using `FULL` is not recommended. For each `UPDATE` or `DELETE` statement, PostgreSQL reads the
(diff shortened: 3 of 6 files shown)


migrate/live-migration/index.md
```diff
@@ -27,20 +26,23 @@ Roughly, it consists of four steps:
...
   inserts are batched. If your application exceeds this, you should use
   the [dual-write and backfill] migration method.
 
-For more information, consult the step-by-step migration guide:
+For more information, refer to the step-by-step migration guide:
 
 - [Live migration from PostgreSQL][from-postgres]
+- [Live migration from TimescaleDB][from-timescaledb]
+
+If you are migrating from AWS RDS to Timescale, you can refer to [this][live-migration-playbook] playbook
```
alejandrodnm, line 41: If they are migrating from RDS then they shouldn't have timescale installed.

alejandrodnm:
Do we want to remove the manual instructions? I thought we were going to leave them for customers with particular use cases that are not served by the image. Maybe they need to do some extra steps, and they can leverage the manual instructions.

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,153 @@
...
+
+```sh
+docker run --rm -it \
+  -e PGCOPYDB_SOURCE_PGURI=$SOURCE \
+  -e PGCOPYDB_TARGET_PGURI=$TARGET \
+  timescale/live-migration:0.0.1
+```
+
+This command may take a while to complete. It’s recommended to run it in the background
+or use terminal multiplexers like `screen` or `tmux`.
```
alejandrodnm: Wouldn't it be better to run it as a daemon?

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,153 @@
...
+```sh
+Migration successfully completed
+```
+
+## 4. Validate the data in target database and promote as new primary
+
+Now that all data has been migrated, the contents of both databases should be the
+same. How exactly this should best be validated is dependent on your application.
+You could compare the number of rows or an aggregate of columns to validate that
+the target database matches with the source.
```
alejandrodnm: This 2 sections talk about validation. I still think that the wait for confirmation to check the integrity is not necessary, specially since we are asking them to validate again when the migration ends.

My gripe is that we leave it for the customer to find out what validation is, which might add more stress to the process (specially since it's on the downtime phase).

What about adding telling them to validate while the data is being replicated, like run some queries with an upper bound that's shared between target and source while the migration is going. You can re-run them again at the end, using the rest of the time.

alejandrodnm: @jgpruitt what did you use to do for validation when doing replication? Did you just assume all the data was there? run some queries?

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,153 @@
...
+
+<Highlight type="note">
+Application downtime begins here.
+</Highlight>
+
+When you're _ready to take downtime_ (i.e., stop normal operations) on your production
+applications that are dependent on your source database and the _lag between the databases is below 30 megabytes_,
+you should stop your applications that perform DML queries on your source database.
+This is the downtime phase and will last until you have completed the validation step (4).
+Be sure to go through the validation step (4) and prepare the validation queries before
```
alejandrodnm: What are the `validation queries`? If we are going to mention something like this, we should explain it further, maybe some examples.

harkishen: Validation queries is a generic term that has been described in step 4. The point here is to basically give a hint to go through the step 4 before continuing here just so that the user is prepared. The preparation is to keep the downtime duration as low as possible. We do mention about some examples in step 4, like `You could compare the number of rows or an aggregate of columns to validate that the target database matches with the source.`

harkishen:
> Do we want to remove the manual instructions? I thought we were going to leave them for customers with particular use cases that are not served by the image. Maybe they need to do some extra steps, and they can leverage the manual instructions.

I am unclear of what advantages will it give. If users want customisation, can those customisations be done post migration? Can those customisations be done when the script pauses for `create_hypertable()`?

My understanding is that we want to give a seamless migration experience, hence keeping 1 doc for Postgres to Timescale & Timescale to Timescale will be good.

migrate/live-migration/live-migration-from-timescaledb.md
harkishen: If you mean as a background process, then this is just a recommendation. I think users may use anything they like.

migrate/live-migration/live-migration-from-timescaledb.md
harkishen: > My gripe is that we leave it for the customer to find out what validation is, which might add more stress to the process (specially since it's on the downtime phase).

I understand the point and agree to it. I want to add that we probably need a tool for verifying data across different databases, because this problem affects other migration approaches as well. Users using pg_dump & pg_restore, or dual-writes backfill will also face a same situation.

> What about adding telling them to validate while the data is being replicated, like run some queries with an upper bound that's shared between target and source while the migration is going.

This may not be right, because there are chances for DML in compressed chunks/old data that will show up as momentary inconsistencies (until it's been streamed to target).

harkishen: > I still think that the wait for confirmation to check the integrity is not necessary, specially since we are asking them to validate again when the migration ends.

Hmm, giving this another thought — we basically ask for 2 signals post starting live migration. They are `s` and `c` key press events.

`s` is asked because
1. We want to know at what moment is the user ready to perform the final step of migration (the step that migrates caggs from source to target). This is important as after this signal, downtime starts.
2. We give user a chance to check if initial data + live-replay data migration matches with source or not. This is mostly for safety reasons based on what I recall.

`c` is asked mainly to continue the remaining steps in migration process. These include copying sequences, caggs, and related work.

Now, the time lapse between receiving `c` and completing the migration is very small.

So, I do agree we can remove the part of validation between `s` and `c` since the time between `c` to end is relatively negligible, so if the data is not consistent, users will know it after the end of migration. Asking them to do the validation few minutes earlier (between `s` & `c`) does not make much sense.


How about this — after live migration starts, we just accept `c` only. Users will provide `c` key press when the _delay between source and target is 0_ and the _user is ready to take downtime_, meaning, we can stop live-replay and proceed with copying sequences, caggs, and remaining work.

@alejandrodnm @arajkumar WDYT?

migrate/live-migration/live-migration-from-timescaledb.md
arajkumar: @alejandrodnm It is a good question. Do you mean docker daemon or system service?

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
-[pgcopydb][pgcopydb] to perform a live migration from a source database which
-is using PostgreSQL to Timescale.
+This document provides detailed instructions on how to perform migration
+from your production PostgreSQL database to a Timescale instance in a way
+that requires minimal downtime (in the order of few minutes) on your
+production applications.
+
+The migration process will be carried out using the [live migration] strategy. You
+will use a live migration image that streamlines the process of replicating live
+transactions from your source database to Timescale.
```
JamesGuthrie:
```suggestion
This document provides detailed instructions to migrate data from your
PostgreSQL database to a Timescale instance with minimal downtime (on the order
of a few minutes) of your production applications, using the [live migration]
strategy. To simplify the migration, we provide you with a docker image
containing all the tools and scripts that you need to perform the live
migration.
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
-treating the data in a snapshot as outdated. If this value is set other than
-`-1`, it might affect the existing data migration step.
+<LiveMigrationStep2 />
 
-To check the current value of `old_snapshot_threshold`, run the command:
-
-```sh
-psql -X -d $SOURCE -c 'show old_snapshot_threshold'
-```
+## 3. Run the live migration image
```
JamesGuthrie:
```suggestion
## 3. Run the live migration docker image
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
 ```
 
-Otherwise, if you are using a managed service, use your cloud provider's
-configuration mechanism to set `old_snapshot_threshold` to `-1`.
-
-Next, you should set `wal_level` to `logical` so that the write-ahead log (WAL)
-records information that is needed for logical decoding.
-
-To check the current value of  `wal_level`, run the command:
+Next, download and run the live migration image:
```
JamesGuthrie:
```suggestion
Next, download and run the live migration docker image:
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
 
 ```sh
-psql -X -d $SOURCE -c 'show wal_level'
+docker run --rm -it \
+  -e PGCOPYDB_SOURCE_PGURI=$SOURCE \
+  -e PGCOPYDB_TARGET_PGURI=$TARGET \
+  timescale/live-migration:0.0.1
```
JamesGuthrie: The suggested tag doesn't exist.
```suggestion
  timescale/live-migration:main
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
 ```
 
-If the query returns something other than `logical`, you must change it.
+This command may take a while to complete. It’s recommended to run it in the
+background or use terminal multiplexers like `screen` or `tmux`.
 
-If you have a superuser on a self-hosted database, run the following command:
+The command will take a snapshot of your source database and migrate the schema
+the target database. Post migrating the schema, it will prompt to create
```
JamesGuthrie:
```suggestion
the target database. After migrating the schema, it will prompt you to create
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
+hypertables in the target database.
 
 ```sh
```
JamesGuthrie: Stray code block
```suggestion
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
-
-```sh
-pgcopydb follow \
-  --source "$SOURCE" \
-  --target "$TARGET" \
-  --fail-fast \
-  --plugin wal2json > pgcopydb_follow.log 2>&1 &
+Execute the following command for each table you want to convert on the target DB:
+    `SELECT create_hypertable('<table name>', '<time column name>', chunk_time_interval => INTERVAL '<chunk interval>')`
 ```
```
JamesGuthrie: stray
```suggestion

```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
-longer the buffered files need to be stored.
-
-## 4. Migrate roles and schema from source to target
+Ideally, tables that contain time-series data should be converted to Hypertables.
+You need to run `create_hypertable()` for each table that you want to convert to
+a Hypertable in the target database. For more information, see [Hypertable docs].
 
-### 4a. Dump the database roles from the source database
+Once you have finished creating Hypertables, you need to signal "continue" to proceed.
+You can do it either by pressing `c` key or giving a `SIGUSR1` signal,
```
JamesGuthrie: I find it confusing that we provide both options "pressing `c` or sending `SIGUSR1`. Why do we have both options? Can we just skip mentioning `SIGUSR1`?

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
-EOF
-```
+When you’re _ready to take downtime_ (i.e., stop normal operations) on your
+production applications that are dependent on your source database and the
+_lag between the databases is below 30 megabytes_, you should stop your
+applications that perform DML queries on your source database. This is
+the downtime phase and will last until you have completed the validation
+step (4). Be sure to go through the validation step (4) and prepare the
+validation queries before you enter the downtime phase to keep the overall
+downtime minimal.
```
JamesGuthrie:
```suggestion
Once the lag between the databases is below 30 megabytes, and you're ready to
take your applications offline, stop all applications which are writing to the
source database. This is the downtime phase and will last until you have
completed the validation step (4). Be sure to go through the validation step
(4) and prepare the validation queries before you enter the downtime phase to
keep the overall downtime minimal.
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
 
-- `replay_lag`: Time elapsed between flushing recent WAL locally and receiving
-  notification that this standby server has written, flushed and applied it.
+Stopping DML queries on your source database allows the live migration
+image to make the target database even with the source. This will be
+evident when the lag outputs 0 megabytes.
```
JamesGuthrie:
```suggestion
Stopping writes to the source database allows the live migration process to
finish replicating data to the target database. This will be evident when the
replication lag reduces to 0 megabytes.
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
-
-Before you proceed, ensure that you have checked the current value of
-`replay_lag`, and that you are comfortable with a partial downtime of this
-duration.
-
-Update the table statistics by running `ANALYZE` on all data:
+When the lag between the two databases shows 0 megabytes, wait for a
+few minutes and then provide the signal to stop the live replay by
+pressing key `s` or giving a `SIGUSR1` signal, as per your log.
+You should see `Stopped` message.
```
JamesGuthrie:
```suggestion
Once the replication lag is 0, wait for a few minutes and then provide the
signal to stop the live replay by pressing key `s` or giving a `SIGUSR1`
signal, as per your log.
You should see a `Stopped` message.
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
-  --resume \
-  --not-consistent
+Migration successfully completed
 ```
 
-Wait until the `pgcopydb follow` process from [step 3][step-3] finishes its
-execution. The process runs until the end position is reached. If you started
-`pgcopydb follow` in the background you can bring it to the foreground with the
-`fg` command.
+## 4. Validate the data in target database and promote as new primary
```
JamesGuthrie: I don't really like the word "promote" here, it implies doing something to the database, which we're not doing.
```suggestion
## 4. Validate the data in target database and use it as new primary
```

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,158 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
-Finally, and clean up the `pgcopydb` artifacts:
-
-```sh
-pgcopydb stream cleanup \
-  --source "$SOURCE" \
-  --target "$TARGET"
-```
+Once you are confident with the validation perspective, the final step is
+to promote the target database as your new production database by using
+the target database's URI in your production application.
```
JamesGuthrie:
```suggestion
Once you are confident with the data validation, the final step is to configure
your applications to use the target database.
```

JamesGuthrie (changes requested):
Initial feedback, still reviewing.

migrate/live-migration/index.md
```diff
@@ -10,10 +10,9 @@ import SourceTargetNote from "versionContent/_partials/_migrate_source_target_no
 # Live migration
 
 Live migration is a migration strategy to move a large amount of data
-(100&nbsp;GB-10&nbsp;TB+) with low downtime (on the order of minutes of
-downtime). It is significantly more complicated to execute than a migration
-with downtime using [pg_dump/restore][pg-dump-and-restore], but supports more
-use-cases and has less requirements than the [dual-write and backfill] method.
+(100&nbsp;GB-10&nbsp;TB+) with low downtime (in the order of few minutes). It requires
```
JamesGuthrie:
```suggestion
(100&nbsp;GB-10&nbsp;TB+) with low downtime (on the order of few minutes). It requires
```

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,153 @@
...
+# Live migration from TimescaleDB database with pgcopydb
+
+This document provides detailed instructions on how to perform a database migration
+from your production TimescaleDB that is either a [Managed Service for TimescaleDB] or
+self-hosted instance to Timescale in a way that requires minimal downtime
+(in the order of few minutes) on your production applications.
+
+The migration process will be carried out using the [live migration] strategy. You
+will use a live migration image that streamlines the process of replicating live
+transactions from your source database to Timescale.
```
JamesGuthrie:
```suggestion
This document provides detailed instructions to migrate data from your
TimescaleDB database (self-hosted or on [Managed Service for TimescaleDB]) to a
Timescale instance with minimal downtime (on the order of a few minutes) of
your production applications, using the [live migration] strategy. To simplify
the migration, we provide you with a docker image containing all the tools and
scripts that you need to perform the live migration.
```

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,153 @@
...
+export TARGET=""
+```
+
+Next, download and run the live migration image:
+
+```sh
+docker run --rm -it \
+  -e PGCOPYDB_SOURCE_PGURI=$SOURCE \
+  -e PGCOPYDB_TARGET_PGURI=$TARGET \
+  timescale/live-migration:0.0.1
```
JamesGuthrie: The suggested tag doesn't exist
```suggestion
  timescale/live-migration:main
```

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,153 @@
...
+
+```sh
+export SOURCE=""
+export TARGET=""
+```
+
+Next, download and run the live migration image:
+
+```sh
+docker run --rm -it \
```
JamesGuthrie: Do we not also need volume mapping to support resume?

JamesGuthrie (changes requested):

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,153 @@
...
+(in the order of few minutes) on your production applications.
+
+The migration process will be carried out using the [live migration] strategy. You
+will use a live migration image that streamlines the process of replicating live
+transactions from your source database to Timescale.
+
+You should provision a dedicated instance to run the migration steps from.
+Ideally an AWS EC2 instance that's in the same region as the Timescale target service.
+For an ingestion load of 10,000 transactions/s, and assuming that the historical
+data copy takes 2 days, we recommend 4 CPUs with 4 to 8 GiB of RAM and 1.2 TiB of storage.
```
arajkumar, line 29: I don't think we still need to ask for 1.2 TiB as we have cleanup scripts in the background?

migrate/live-migration/live-migration-from-timescaledb.md
JamesGuthrie: Since we're already using docker, I would propose that we tell users to start the container with `-dit` (detached interactive tty), and that they then use `docker attach <container name>` to attach to the container. That will decouple their shell session from the migration tool.

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,153 @@
...
+
+In detail, the migration process consists of the following steps:
+1. Set up a target database instance in Timescale.
+1. Prepare the source database for live migration.
+1. Run the live migration image.
+1. Validate the data in target database and promote as new primary.
+
+<GettingHelp />
+
+<StepOne />
```
arajkumar, line 41: Not necessarily in this PR, but we should consider adding steps on how to migrate extension from SOURCE to TARGET.

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,153 @@
...
+```sh
+[WATCH] Source DB - Target DB => 126MB
+```
+
+When the lag between the source and target database is less than 30 megabytes, it will
+start `ANALYZE` on the target database. This updates statistics in the target database,
+which is necessary for optimal querying performance in the target database. Wait for
+`ANALYZE` to complete.
+
+<Highlight type="note">
```
arajkumar: Should the type be "warning"?

migrate/live-migration/live-migration-from-postgres.md
arajkumar: +1, since we advocate to use docker in interactive mode(-it), there is no need to mention about signals.

migrate/live-migration/live-migration-from-timescaledb.md
arajkumar: +1. Additionally I recommend to pass `--name live-migration`.

migrate/live-migration/live-migration-from-timescaledb.md
harkishen: @arajkumar and I discussed about this, and we agreed that we should allow users to check data consistency at the end of migration process. ATM, user have to validate data twice — first, when he gives `s`, he checks for existing data + live replay, but without caggs (since they are yet to be migrated); second time, after migration completes, he verifies for the entire database. With the new approach, user will only perform verification on entire database after the migration completes. Hence, we will keep `c`, and remove the `s` signal.

For instance, the new workflow will be as follows.

User starts live migration for TS to TS. When the delay between source and target is 0 (he may have to stop load on the source side) and the user is ready to take the "minimal downtime", he will press `c` key. The script will then stop live-replay and do the remaining work to complete the migration process. 

If the migration is for PG to TS, an additional step for `create_hypertable()` will be asked before the `c` key press.

migrate/live-migration/live-migration-from-timescaledb.md
alejandrodnm: Yeah, I meant detached `-d`

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,150 @@
...
+
+<Highlight type="important">
+Application downtime begins here.
+</Highlight>
+
+Once the lag between the databases is below 30 megabytes, and you're ready to
+take your applications offline, stop all applications which are writing to the
+source database. This is the downtime phase and will last until you have
+completed the validation step (4). Be sure to go through the validation step
+(4) and prepare the validation queries before you enter the downtime phase to
```
alejandrodnm: Again, I think the phrase

> prepare the validation queries before you enter the downtime phase

Is confusing. If I had zero context I would imagine there are some validation scripts somewhere in the instructions.

If we want the customer to have something prepare for validation, we need to specify at the beginning of the instructions, not at the end. It could be an optional step of 'Prepare validation scripts'.

alejandrodnm:
> > Do we want to remove the manual instructions? I thought we were going to leave them for customers with particular use cases that are not served by the image. Maybe they need to do some extra steps, and they can leverage the manual instructions.
> 
> I am unclear of what advantages will it give. If users want customisation, can those customisations be done post migration? Can those customisations be done when the script pauses for `create_hypertable()`?
> 
> My understanding is that we want to give a seamless migration experience, hence keeping 1 doc for Postgres to Timescale & Timescale to Timescale will be good.

That's the thing, we don't know what crazy things they might ask, they might want to exclude some tables or whatever. We can always say, if the tool doesn't fit the use case you still have the manual instructions.

migrate/live-migration/live-migration-from-timescaledb.md
harkishen: Update: With https://github.com/timescale/pgcopydb/pull/22 merged, the live migration image behaviour aligns with the above mentioned approach.

migrate/live-migration/live-migration-from-timescaledb.md
harkishen: Yes, we have cleanup scripts in background, but I don't have numbers with cleanup script. However, if the data copy takes 2 days to complete, maybe its worth having 1.2TBs?

migrate/live-migration/live-migration-from-timescaledb.md
harkishen: Since we don't have validation scripts ATM, and they are creating confusion, how about we remove the part ` and prepare the validation queries ...`?

alejandrodnm: Let's do that.

migrate/live-migration/live-migration-from-timescaledb.md
harkishen: Based on my understanding, I don't think we should suggest detached mode. If we suggest `-dit`, then we will have to add about attaching the container and reading when to send signal based on the delay. Suggesting simply `screen/tmux` covers it up.

Thoughts?

migrate/live-migration/live-migration-from-postgres.md
```diff
@@ -12,402 +12,130 @@ import StepOne from "versionContent/_partials/_migrate_dual_write_step1.mdx";
...
-to find all matching rows, which will result in significantly slower
-replication.
-
-```sql
-ALTER TABLE {table_name} REPLICA IDENTITY FULL
-```
-
-Once you're sure all your tables are properly configured for replication, use
-`pgcopydb`'s follow command to create a replication slot:
+First, set the database URIs as environment variables:
```
JamesGuthrie, line 73: This instruction is too late. The previous steps refer to `$SOURCE`. Also, I would prefer it if we could provide an actual url as an example, instead of an empty string. Compare with the instructions in the note for dual-write and backfill: https://docs.timescale.com/migrate/latest/dual-write-and-backfill/dual-write-from-timescaledb/#3-set-up-schema-and-migrate-relational-data-to-target-database

image

migrate/live-migration/live-migration-from-timescaledb.md
JamesGuthrie: @Harkishen-Singh This change hasn't been made yet, or am I missing something?

JamesGuthrie: > have to add about attaching the container and reading when to send signal based on the delay
> ...
> Suggesting simply screen/tmux covers it up.

But suggesting it doesn't mean that people will do it, or that they know how to do it. Why would we need to give the user instructions on how to do it with Docker, but we don't have to give them instructions if they're using screen/tmux?

_partials/_migrate_live_migration_step2.md
```diff
@@ -0,0 +1,54 @@
+<Highlight type="note">
+For sake of convenience, set `$SOURCE` and `$TARGET` connection strings
+in your migration instance.
+
+```sh
+export SOURCE="postgres://user:password@host:port/database"
+export TARGET="postgres://user:password@host:port/database"
+```
+</Highlight>
```
JamesGuthrie: This duplicated the content which is already in a partial, `./_partials/_migrate_set_up_source_and_target.md` please use that instead.

migrate/live-migration/live-migration-from-timescaledb.md
alejandrodnm: BTW I've just realize you can use `docker run` and detach with CTRL-P+Ctrl-Q

https://docs.docker.com/engine/reference/commandline/cli/#default-key-sequence-to-detach-from-containers

JamesGuthrie (approved):
I won't be able to give a proper final review, but I don't want to block the merge, so I'm approving.

harkishen:
I will wait for other approvals before merging. We should also probably wait for product for the target launch date before merging.

JamesGuthrie:
I think that we should merge docs as soon as they're ready. @VineethReddy02 WDYT?

migrate/live-migration/live-migration-from-timescaledb.md
arajkumar: Ahh, I forgot about that :D

alejandrodnm (approved):

migrate/live-migration/live-migration-from-timescaledb.md
```diff
@@ -0,0 +1,159 @@
...
+export TARGET=""
+```
+
+Next, download and run the live migration docker image:
+
+```sh
+docker run --rm -dit --name live-migration \
+  -e PGCOPYDB_SOURCE_PGURI=$SOURCE \
+  -e PGCOPYDB_TARGET_PGURI=$TARGET \
+  -v ~/tmp/live-migration:/opt/timescale/ts_cdc \
```
arajkumar: It is better to be explit about the host directory which is being mounted to docker must have enough disk space.

```suggestion
  -v ~/live-migration:/opt/timescale/ts_cdc \
```

VineethReddy02:
@Harkishen-Singh is this PR ready to be merged?

migrate/live-migration/live-migration-from-timescaledb.md
harkishen: I have also added a note below this line

```
Ensure `/live-migration` directory has sufficient space available.
Ideally, 1.5 times the source database size should be good.
```

harkishen:
@VineethReddy02 yes.

arajkumar (approved):

PR #2970, opened 2024-01-31
Add pg_dumpall error note

alejandrodnm:
# Description

When running pg_dumpall from MST or RDS some additional flags are required otherwise the command will throw an error:

- `--no-role-password` will not dump the passwords for the roles, this solves the `permission denied for table pg_authid` error. Since it means an extra step for the customer, they would have to manually set the passwords, we put it as an important note. Let them try first to see if it works and they can dump the password.
- `-l DB_NAME` solves the connection to `template1` error. Since this won't incur in additional steps for the user, we always set it up in the instructions.

# Links

Fixes https://github.com/timescale/team-data-onboarding/issues/8

### DIFF
--- a/_partials/_migrate_dual_write_dump_database_roles.md
+++ b/_partials/_migrate_dual_write_dump_database_roles.md
@@ -1,9 +1,23 @@
 ```bash
 pg_dumpall -d "$SOURCE" \
+  -l $DB_NAME \
   --quote-all-identifiers \
   --roles-only \
   --file=roles.sql
 ```
+
+<Highlight type="important">
+Some providers like Managed Service for TimescaleDB (MST) and AWS RDS don't
+support role password dumps. If dumping the passwords results in the error:
+
+```
+pg_dumpall: error: query failed: ERROR:  permission denied for table pg_authid
+```
+
+Execute the command adding the `--no-role-passwords` flag. After restoring the
+roles into the target database, manually set passwords with `ALTER ROLE name
+WITH PASSWORD '<YOUR_PASSOWRD>';`
+</Highlight>
 
 Timescale services do not support roles with superuser access. If your SQL
 dump includes roles that have such permissions, you'll need to modify the file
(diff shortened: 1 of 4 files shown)


_partials/_migrate_dual_write_dump_database_roles.md
```diff
@@ -1,10 +1,23 @@
...
 
+<Highlight type="important">
+Some providers like Managed Service for TimescaleDB (MST) and AWS RDS don't
+support role password dumps. If dumping the passwords results in the error:
+
+```
+pg_dumpall: error: query failed: ERROR:  permission denied for table pg_authid
+```
+
+Execute the command adding the `--no-role-passwords` flag. Then, manually set
```
JamesGuthrie: I think that we should be more specific about when the passwords should be set.

I assume that it would be something like:

> After restoring the roles into the target database, you must manually set...

JamesGuthrie (changes requested):

_partials/_migrate_dual_write_dump_database_roles.md
alejandrodnm: I updated it.

_partials/_migrate_set_up_source_and_target.md
```diff
@@ -5,7 +5,7 @@ databases are referred to as `$SOURCE` and `$TARGET` throughout this guide.
 This can be set in your shell, for example:
 
 ```bash
-export SOURCE=postgres://<user>:<password>@<source host>:<source port>/<dbname>
-export TARGET=postgres://<user>:<password>@<target host>:<target port>/<dbname>
+export SOURCE=postgres://<user>:<password>@<source host>:<source port>/<db_name>
```
alejandrodnm, line 8: I changed this, because it's places before the `pg_dumpall` instructions that have the `-l $DB_NAME`. That way it's easier to make the connection when reading the instructions.

JamesGuthrie (approved):

PR #3030, opened 2024-02-26
Live migration is more complex than pg_dump/restore

JamesGuthrie:
# Description

Live migration is more complex than pg_dump/restore

### DIFF
--- a/migrate/live-migration/index.md
+++ b/migrate/live-migration/index.md
@@ -12,7 +12,7 @@ import OpenSupportRequest from "versionContent/_partials/_migrate_open_support_r
 
 Live migration is a new migration strategy to move a large amount of data
 (100&nbsp;GB-10&nbsp;TB+) with low downtime (on the order of few minutes). It
-requires more steps to execute than a migration with downtime using
+is significantly more complex to execute than a migration with downtime using
 [pg_dump/restore][pg-dump-and-restore], but supports more use-cases and has
 fewer requirements than the [dual-write and backfill] method.
 


migrate/live-migration/index.md
```diff
@@ -12,7 +12,7 @@ import OpenSupportRequest from "versionContent/_partials/_migrate_open_support_r
 
 Live migration is a new migration strategy to move a large amount of data
 (100&nbsp;GB-10&nbsp;TB+) with low downtime (on the order of few minutes). It
-requires more steps to execute than a migration with downtime using
+is significantly more complex to execute than a migration with downtime using
```
VineethReddy02, line 15: The introduction of a Docker image has simplified the process of live migration. The only wrinkle is stopping the CDC once its in sync.

We already have banners in our documentation, stating that live migrations are in an experimental or public preview stage. Thus, I suggest we remove the mention of “significantly complex” as discourages the user to consider this option.

VineethReddy02 (changes requested):

migrate/live-migration/index.md
VineethReddy02: Perhaps we should eliminate the comparison with dump/restore in terms of ease of use, as the value they provide is completely different. It's an unfair comparison. 🙂

harkishen: How about phrasing it gently as "can require more work to execute than a migration"?

JamesGuthrie: The reality is that live migration _is_ significantly more complicated than pg_dump/restore. Users should only use it _if they really need to_. IMHO we should be clear about this.

arajkumar (approved):

VineethReddy02 (approved):

migrate/live-migration/index.md
VineethReddy02: I'm not completely convinced, but let's do this.

PR #3050, opened 2024-03-05
fix: tweak live migration disk recommendation

JamesGuthrie:
# Description

Live migration doesn't need intermediate storage for the initial data copy, so the disk requirements are lower.


arajkumar (approved):

harkishen (approved):
Looks good, but something to note; the changes made here belong to a `<Highlight>` block. I believe we can move the entire highlight block to partials.

PR #3162, opened 2024-04-25
Add warning about SERIAL types and dual-write

JamesGuthrie:
# Description

Add warning about SERIAL types and dual-write


_partials/_migrate_dual_write_step2.md
```diff
@@ -7,8 +7,20 @@ you must think about how to handle the failure to write to either the source or
 target database, and what mechanism you want to or can build to recover from
 such a failure.
 
+Should your time-series data have foreign-key references into a plain table,
+you must ensure that your application correctly maintains the foreign key
+relations. Beware that if the referenced column is a `*SERIAL` type, the same
```
billy-the-fish: Can you change Beware that if to simple If please.

_partials/_migrate_dual_write_step2.md
```diff
@@ -7,8 +7,20 @@ you must think about how to handle the failure to write to either the source or
...
+row inserted into the source and target _may not_ obtain the same autogenerated
+id. If this happens, the data backfilled from the source to the target will be
```
billy-the-fish: Always use the present tense:

from the source to the target is

_partials/_migrate_dual_write_step2.md
```diff
@@ -7,8 +7,20 @@ you must think about how to handle the failure to write to either the source or
...
+internally corrupt. In the best case it causes a foreign key violation, in the
+worst case, the foreign key relation constraint is maintained, but the data
+references the wrong foreign key. If reading this has made you worried,
+consider using [live migration] instead.
```
billy-the-fish: If reading this has made you worried, consider using [live migration] instead.

To avoid these issues, best practice is to follow [live migration][live-migration].

billy-the-fish (changes requested):
Nice doc. Couple of small language updates for you.

_partials/_migrate_dual_write_step2.md
JamesGuthrie: done

_partials/_migrate_dual_write_step2.md
JamesGuthrie: done

_partials/_migrate_dual_write_step2.md
JamesGuthrie: done

billy-the-fish (approved):
LGTM

_partials/_migrate_dual_write_step2.md
```diff
@@ -7,8 +7,20 @@ you must think about how to handle the failure to write to either the source or
 target database, and what mechanism you want to or can build to recover from
 such a failure.
 
+Should your time-series data have foreign-key references into a plain table,
```
alejandrodnm, line 10: I fail to see when something like this could happen. When the non-TS data is migrated then all the IDs that are going to be referenced by the backfilled data should be already present in TARGET. No?

JamesGuthrie: _If_ the customer is _only_ dual-writing timeseries data, yes. If they intend to dual-write _both_ the referenced data and the timeseries data then this can happen.

PR #3253, opened 2024-06-14
fix: numbering and first docker connection.

billy-the-fish:


_partials/_install-self-hosted-docker-based.mdx
```diff
@@ -19,15 +18,13 @@
 
     If you are running multiple container instances remember to change the port each Docker instance runs on.
 
-
  1. **Connect to a database on your PostgreSQL instance**
```
xenophenes, line 21: Aren't these the numbers that need to be adjusted?

_partials/_install-self-hosted-docker-based.mdx
```diff
@@ -19,15 +18,13 @@
...
 
-   In PostgreSQL, the default user and database are both `postgres`. To use a
-   different database, set `<database-name>` to the name of that database:
+   In PostgreSQL, the default user and database are both `postgres`. The default command for this Docker container is:
 
-   ```bash
-   psql -d "postgres://<username>:<password>@<host>:<port>/<database-name>"
-   ```
+    ```bash
+    psql -d "postgres://postgres:password@localhost/postgres"
```
xenophenes, line 31: Is the password literally "password"? I can't just copy and paste this command, so something else is missing. (`psql: error: connection to server at "localhost" (::1), port 5432 failed: FATAL:  role "postgres" does not exist`)

_partials/_install-self-hosted-docker-based.mdx
billy-the-fish: Gatsby does autonumbering. Best practice is to always use 1. This went a bit strange before because I left 2 newlines between the numbers.

_partials/_install-self-hosted-docker-based.mdx
billy-the-fish: If you run the command in the previous step for the *pg16 image this works (on my machine :-( ). I have added an explanation.

xenophenes (approved):

PR #3316, opened 2024-07-12
chore: changelog

billy-the-fish:
Add the latest improvements in Timescale products to the changelog,


willietran (changes requested):
Per Kostis' message in Slack, there's a chance this hasn't been released to the public yet. I think it'd be good to remove the Live Migrations section for now.

willietran (approved):

PR #3387, opened 2024-08-20
Minor nits in migration documentation

JamesGuthrie:


_partials/_migrate_awsrds_migrate_data_downtime.md
```diff
@@ -88,11 +88,10 @@ service:
      --no-owner \
      --no-privileges \
      --section=pre-data \
-     --file=pre-data-dump.sql \
-     --snapshot=$(cat /tmp/pgcopydb/snapshot)
+     --file=pre-data-dump.sql
    ```
```
arajkumar: I don't think we need to split the migration into pre-data and post-data. It should look very similar to Postgres to Timescale (i.e just dump and restore)

arajkumar (changes requested):

_partials/_migrate_awsrds_migrate_data_downtime.md
JamesGuthrie: The old RDS to Timescale dump/restore had this split, because between pre-data and post-data it prompted the user to convert tables to hypertables (I assume to prevent having to copy the data into a plain table and then re-copy when the table was converted to a hypertable).

It looks like we removed this split n the Postgres -> Timescale migration, so I will also remove it here.

arajkumar (approved):

PR #3401, opened 2024-08-29
fix parameter name

FrVaBe:
# Description

On the [Use a user-defined action to implement automatic tablespace management](https://docs.timescale.com/use-timescale/latest/user-defined-actions/example-tiered-storage/#use-a-user-defined-action-to-implement-automatic-tablespace-management) documentation site the parameter name in the add_job Statement does not match the parameter name in the move_chunks procedure. Thus the validation of required parameters in the move_chunks procedure will always fail.
This is fixed now.

# Links

Did not fill an issue.

### DIFF
--- a/use-timescale/user-defined-actions/example-tiered-storage.md
+++ b/use-timescale/user-defined-actions/example-tiered-storage.md
@@ -82,7 +82,7 @@ to move data to low-cost object storage backed by Amazon S3.
     SELECT add_job(
       'move_chunks',
       '1d',
-      config => '{"hypertable":"metrics","lag":"12 month","tablespace":"old_chunks"}'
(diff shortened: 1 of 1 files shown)


jonatas (approved):
Thanks for the fix @FrVaBe 🙌

FrVaBe:
@jonatas No thanks at all. By the way I realized that the Note at the end of the page about 
> This procedure uses PostgreSQL's regular ALTER TABLE ... SET TABLESPACE syntax

is also obsolete because the example does not use the  ALTER TABLE ... SET TABLESPACE syntax anymore but already the `move_chunks ` function.  Can you fix this quickly or should I create another pull request?

jonatas:
Thanks again @FrVaBe! Removed the message on #3408.

PR #3433, opened 2024-09-13
Reference that continuous aggregates can be tiered in intro

mfreed:
Reference that continuous aggregates can be tiered in introduction section to tiering, as well as other operations that work transparently with tiering.


billy-the-fish (approved):
LGTM

PR #3446, opened 2024-09-17
fix: update docs about partitioning.

billy-the-fish:


api/add_dimension.md
```diff
@@ -11,79 +11,52 @@ api:
...
+  Best practice is to use RAID when possible. This is because it supports the concurrent and single
+  scenarios and *no spatial partitioning is required*.
+
+- **Multiple tablespaces**: for each physical disk, add a separate tablespace to the database. TimescaleDB allows you to add 
+  multiple tablespaces to a *single* hypertable. However, although under the hood, a hypertable's
+  chunks are spread across the tablespaces associated with that hypertable.
+
+  Multiple tablespaces only supports concurrent queries.
+
+When using hash partitions, best practice is to use 1 hash partition per disk.
```
erimatnor: I am not sure we have a best practice for this. But I guess it is reasonably to have _at least_ 1 partition per disk. Personally, I would probably say it is better to have number of partitions set to a multiple of number of disks, e.g., the number of partitions P=N*Pd where N is the number of disks and Pd is the number of partitions per disk. Then you could also add more disks later and move partitions to the new disk from other disks. But maybe this is too much information for this section?

api/add_dimension.md
```diff
@@ -119,7 +92,7 @@ partitioning on column `time`, then add an additional partition key on
 
 ```sql
 SELECT create_hypertable('conditions', by_range('time'));
-SELECT add_dimension('conditions', by_hash('location', 4));
+SELECT add_dimension('conditions', by_range('time'), by_hash('location', 4));
```
erimatnor: I don't think this change is correct. Looks like you are confusing `create_hypertable` and `add_dimension`.

`create_hypertable` requires adding at least one range dimension as the "primary" partitioning dimension. However, `add_dimension` is called _after_ `create_hypertable` to add additional dimensions, typically a space/hash dimension after the hypertable exists  (including its primary range/time dimension).

api/add_dimension.md
```diff
@@ -133,13 +106,13 @@ partitionining on `device_id`.
 
 ```sql
 SELECT create_hypertable('conditions', by_range('time'));
-SELECT add_dimension('conditions', , by_hash('location', 2));
+SELECT add_dimension('conditions', by_range('time'), by_hash('location', 2));
```
erimatnor: Same here as above. And, the changes a few lines below.

api/dimension_info.md
```diff
@@ -1,41 +1,36 @@
...
-Dimension builders were introduced in TimescaleDB 2.13.
-</Highlight>
-
-The `create_hypertable` and `add_dimension` are used together with
-dimension builders to specify the dimensions to partition a
-hypertable on.
+You call [`create_hypertable`][create_hypertable] and [`add_dimension`][add_dimension] to specify the dimensions to 
+partition a hypertable on. TimescaleDB supports partitioning [`by_range`][by-range] and [`by_hash`][by-hash]. You can 
+partition `by_range` on it's own. However, because `by_hash` makes a fixed number of partitions which can grow 
+very large, best practice is to set `by_range` when you use `by_hash`.
```
erimatnor: Here I suggest we say that hypertables must always have a primary range/time dimension, followed by an arbitrary number of additional dimensions that can be either range or hash, but typically just one hash.

api/add_dimension.md
billy-the-fish: Updated.

api/add_dimension.md
billy-the-fish: Updated.

api/add_dimension.md
billy-the-fish: Updated.

api/dimension_info.md
billy-the-fish: Done

api/dimension_info.md
```diff
@@ -1,41 +1,39 @@
...
+ 
 
-<Highlight type="tip">
-For incompatible data types (for example, `jsonb`) you can specify a function to
+For incompatible data types such as `jsonb`, you can specify a function to
 the `partition_func` argument of the dimension build to extract a compatible
 data type. Look in the example section below.
-</Highlight>
+
+Dimension builders were introduced in TimescaleDB 2.13.
```
erimatnor: I am confused by the section on "dimension builders". I know it was in the docs before, but it seems the concept is never really introduced. What is a dimension builder and why do I care about it?

My understanding is that it is just a set of functions you use together with create_hypertable() and add_dimension(), but those pages don't even mention the concept or make any reference to this section although the "builder" functions are used in those sections.  

I guess am wondering why we need a separate section on dimension builders when it is not mentioned elsewhere?

If we are sticking with this concept of dimension builders, I think this section needs a better introduction to what they are and why they are useful, and when to use them.

But, if it was up to me, I would just take what is useful here and merge it with the sections on create_hypertable and add_dimension. I would probably skip the concept of "dimension builder" and just explain how to use the functions by_range() and by_hash()

api/add_dimension.md
```diff
@@ -11,79 +11,62 @@ api:
...
+  That is, using a single tablespace.
+
+  Best practice is to use RAID when possible. This is because it supports the concurrent and single
+  scenarios and *no spatial partitioning is required*.
+
+- **Multiple tablespaces**: for each physical disk, add a separate tablespace to the database. TimescaleDB allows you to add 
+  multiple tablespaces to a *single* hypertable. However, although under the hood, a hypertable's
+  chunks are spread across the tablespaces associated with that hypertable.
+
+  Multiple tablespaces only supports concurrent queries.
```
erimatnor: This isn't correct. A single query should be able to read from multiple tablespaces/disks and also run a parallel query across those disks.

api/add_dimension.md
```diff
@@ -11,79 +11,62 @@ api:
...
+- Two or more concurrent queries should be able to read from different disks in parallel.
+- A single query should be able to use query parallelization to read from multiple disks in parallel.
+
+For the following options:
+
+- **RAID**: use a RAID setup across multiple physical disks, and expose a single logical disk to the hypertable.
+  That is, using a single tablespace.
+
+  Best practice is to use RAID when possible. This is because it supports the concurrent and single
+  scenarios and *no spatial partitioning is required*.
```
erimatnor: This was probably there before and I am not sure where it comes from, but this seems to imply that using multiple tablespaces/disks requires spatial partitioning. This is not the case (and I don't think it has ever been the case). You can spread chunks across multiple disks with only time partitioning, so I am not sure exactly why "no spatial partitioning is required" is mentioned since it applies also the "Multiple tablespaces".

I would recommend just removing this last bit.

If we want to mention that RAID is "best practice" (and, honestly, I am not exactly sure why it is best practice), I would motivate it primarily with the fact that you need not manually manage tablespaces in the DB when using RAID. Still, you do need the complexity of a RAID setup instead.

api/add_dimension.md
```diff
@@ -11,79 +11,62 @@ api:
...
+  multiple tablespaces to a *single* hypertable. However, although under the hood, a hypertable's
+  chunks are spread across the tablespaces associated with that hypertable.
+
+  Multiple tablespaces only supports concurrent queries.
+
+When using hash partitions, best practice is to have at least one hash partition per disk.
+
+set number of partitions to a multiple of number of disks. For example, the number of 
+partitions P=N*Pd where N is the number of disks and Pd is the number of partitions per 
+disk. This enables you to add more disks later and move partitions to the new disk from other disks.
```
erimatnor:
```suggestion
When using multiple tablespaces, a best practice is to also add a second hash-partitioned dimension to your hypertable and to have at least one hash partition per disk. While a single time dimension would also work, it would mean that the first chunk is written to one tablespace, the second to another, and so on, and thus would parallelize only if a query's time range exceeds a single chunk.

When adding a hash partitioned dimension, set the number of partitions to a multiple of number of disks. For example, the number of 
partitions P=N*Pd where N is the number of disks and Pd is the number of partitions per 
disk. This enables you to add more disks later and move partitions to the new disk from other disks.
```

api/dimension_info.md
billy-the-fish: Thanks for this comment, it really helps. I have had another look over this, and from what I understand this is talking about the _timescaledb_internal.dimension_info type returned when you call create_hypertable or add_dimension. How about as you suggest I add any useful information to those functions, and move the rest of the info to https://docs.timescale.com/api/latest/informational-views/ as a dimension_info object? 

My head hurts.

billy-the-fish:
The merge was a mess, replaced with https://github.com/timescale/docs/pull/3608.

PR #3449, opened 2024-09-18
Add a doc on chunk skipping indexes

### LINKED ISSUE #3424: [Docs RFC] Add doc on chunk skipping indexes for CAGGS
We got the work on chunk skipping indexes out with 2.16. As a quick reminder:
> Improved query performance through chunk exclusion on compressed hypertables.
> You can now define chunk skipping indexes on compressed chunks for any column with one of the following integer data types: smallint, int, bigint, serial, bigserial, date, timestamp, timestamptz.
> After you call enable_chunk_skipping on a column, TimescaleDB tracks the min and max values for that column. TimescaleDB uses that information to exclude chunks for queries that filter on that column, and would not find any data in those chunks.

All is great; we have docs (for [enabling](https://docs.timescale.com/api/latest/hypertable/enable_chunk_skipping/) / [disabling](https://docs.timescale.com/api/latest/hypertable/disable_chunk_skipping/)) and an extended [blog post](https://docs.google.com/document/d/1wplgKvAO2DEW6sOdU3JJ1oIgaRGGKJWfXAT4xp6A6Y0/edit) coming out next week.

But I was thinking that we lack any further info on our docs for this pretty useful feature that users have to manually enable. We only have the API pages, with not much information on the why and how this helps, not to mention that someone would have to know about this to look up those commands on our API section.

So I was wondering where we could add more info about it - maybe condense the [blog post](https://docs.google.com/document/d/1wplgKvAO2DEW6sOdU3JJ1oIgaRGGKJWfXAT4xp6A6Y0/edit) somehow (remove intro/outro and all the marketing material) and add it somewhere as a page - not sure if it should be a separate page under "Use Timescale" for example.

atovpeko:


use-timescale/hypertables/improve-query-performance.md
```diff
@@ -0,0 +1,44 @@
+---
+title: Improve query performance
+excerpt: How and why to use chunk skipping to optimize hypertable performance
+products: [cloud, mst, self_hosted]
+keywords: [hypertables, indexes, chunks]
+---
+
+# Improve query performance
```
iroussos, line 8: This is only one of the ways to improve query performance.

@billy-the-fish @atovpeko If this makes sense to you as well, I think that we should have a whole section on how to "Improve query performance", with this page just being one of the many that we'll add.

Other candidates:
- Enable Compression - With information on how to setup compression correctly (especially segment_by, order_by and the differences between asc/desc options on the later). Also including information on our compression recommendations (also still missing from the docs I think @cevian?)
- Add more indexes - With information on indexes, but also how we add min/max sparse indexes on compressed chunks when we see an index on the main hypertable (more to come on this - we'll write a blog post soon that we'll be able to use)
- Add continuous aggregates - With an intro on why and how CAggs are useful and how to re-write queries to use them. 
  [Advanced] How and why setup hierarchical continuous aggregates
- [Advanced] Setting (and updating) the chunk size for hypertables
  [Even more advanced] Setting (and updating) the chunk size for Continuous Aggregates
- Fine tune policies - Not exactly related to query performance, but still pretty relevant
- _I bet there is something more that I may be forgetting right now.._ 

Now, back to this - this is about improving the performance of a very specific use case: queries that do not filter by a partitioning column.

Not sure what the best title could be, but the following sounds like too much (but maybe it's ok to get us started):
```suggestion
# Improve performance of queries that do not filter by a partitioning column
```

use-timescale/hypertables/improve-query-performance.md
```diff
@@ -0,0 +1,44 @@
...
+
+When you execute a query on a hypertable, you do not parse the whole table; you only access the chunks necessary
+to satisfy the query. This works well when the `WHERE` clause of a query uses the column by which a hypertable is 
+partitioned. For example, in a hypertable where every day of the year is a separate chunk, a query for September 1 
+accesses only the chunk for that day. However, many queries use columns other than the partitioning one. For example, 
+a table might have two columns: one for when data was gathered and one for when it was added to the database. If you 
+partition by the date of gathering, a query by the date of adding accesses all chunks in the hypertable and slows the 
+performance.
+
```
iroussos, line 17: Why be shy and not add the example and the diagrams from the [blog post](https://www.timescale.com/blog/boost-postgres-performance-by-7x-with-chunk-skipping-indexes/)? 

Same question for bellow on the "How chunk skipping works"

use-timescale/hypertables/improve-query-performance.md
```diff
@@ -0,0 +1,44 @@
...
+you need.
+
+## Enable chunk skipping
+
+To enable chunk skipping on a column, call `enable_chunk_skipping` on a `hypertable` for a `column_name`. For example, 
+the following query enables chunk skipping on the `order_id` column in the `orders` table:
+
+```sql
+SELECT enable_chunk_skipping('orders', 'order_id');
+```
```
iroussos, line 54: We need to make it more explicit that this works only for compressed chunks. 

I just checked and realized that we are missing this info on the [api page](https://docs.timescale.com/api/latest/hypertable/enable_chunk_skipping/) as well - we talk about hypertables in general  at the top and just talk about compressed chunks as a side note at the end :-(

use-timescale/hypertables/improve-query-performance.md
```diff
@@ -0,0 +1,44 @@
...
+
+For more details, see the [API Reference][api-reference].
+
```
iroussos, line 57: I think that we are missing a section on when does this work?

Rough sketch (by copy pasting something from the blog post that may not make sense right now without the example): 

```suggestion

## When should you use chunk-skipping indexes?

So the question is when should you use chunk-skipping indexes and for which columns. 

This works best for columns that are related to the partitioning column in some way; they are correlated. To exemplify this correlation, let’s think of our earlier example of ending a job. A job’s end time follows its creation time, typically by seconds or minutes. Similarly, a satellite observation’s recording time on the database follows the actual observation time. Sequential IDs and the creation timestamp for the same entities also correlate (both increasing synchronously).
```

iroussos (changes requested):
Thank you for adding this page @atovpeko!

I have added a few comments and thoughts - let me know what you think :-)

use-timescale/hypertables/improve-query-performance.md
atovpeko: You are very right in this :-) Iain suggested this page title to be "Improve query performance" because we plan on adding more optimization stuff in there - eventually making it into a section, I think. However, changing titles back and forth won't be good for SEO or doc user experience, that's why we put in this more general title from the very beginning.

use-timescale/hypertables/improve-query-performance.md
atovpeko: Good point on diagrams, thank you. For examples, I was aiming at making this a very lean document (as tech docs should be). I also didn't want this to copy your blog post word for word, for SEO and other reasons. It seemed to me that having more examples would not make the point much clearer (again, because this is a tech document, rules are very different for tech marketing content). I can add more, of course, if you feel that having only one does not convey the point. I think it would be better though to just explicitly link to your blog post for an in-depth look.

atovpeko: I made the example more specific, linked to your blog post, and added a diagram

use-timescale/hypertables/improve-query-performance.md
atovpeko: Added a highlight to the main doc and emphasis in API docs

use-timescale/hypertables/improve-query-performance.md
atovpeko: Added, thank you.

use-timescale/hypertables/improve-query-performance.md
```diff
@@ -0,0 +1,59 @@
...
+specifies ranges on the column. 
+
+![Chunk skipping](https://assets.timescale.com/docs/images/hypertable-with-chunk-skipping.png)
+
+You can enable chunk skipping on compressed hypertables for `smallint`, `int`, `bigint`, `serial`, `bigserial`, `date`,
+`timestamp`, or `timestamptz` type columns. 
+
+## When to enable chunk skipping
+
+You can enable chunk skipping on as many columns as you need. However, best practice is to enable it on the following columns:
```
billy-the-fish:
```suggestion
You can enable chunk skipping on as many columns as you need. However, best practice is to enable it on columns that are both:
```

use-timescale/hypertables/improve-query-performance.md
```diff
@@ -0,0 +1,59 @@
...
+ 
+- correlated, that is, related to the partitioning column in some way, and
```
billy-the-fish:
```suggestion
- Correlated. That is, related to the partitioning column in some way
```

use-timescale/hypertables/improve-query-performance.md
```diff
@@ -0,0 +1,59 @@
...
+- referenced in the `WHERE` clauses.
```
billy-the-fish:
```suggestion
- Referenced in the `WHERE` clauses. 
```

billy-the-fish (approved):
LGTM

billy-the-fish:
@iroussos , are you ok for us to publish this doc?

use-timescale/hypertables/improve-query-performance.md
iroussos: What I  was thinking about was this one being just one of **multiple pages** under a larger section on the docs called "Improved Query Performance", not as a section on a larger page (cause it already feels like it has a lot and we can add way more following my comment above).

But if you think that the way to go is one page for all, I am not going to object - you know better about how to optimally structure our documentation!

iroussos (approved):
Thank you @atovpeko @billy-the-fish! The doc looks great for a first iteration, no reason to hold it any more, we can always come back and refine if there is any need to go deeper :-) 

My only comment is the one I responded again and left unresolved - but yours to think and decide about, approving on my end. 

Once more, thank you for the great work on this!

use-timescale/hypertables/improve-query-performance.md
atovpeko: Yeah, I also meant making it into a section with multiple pages, not a section on the page. We'll get there :-) Thank you!

PR #3512, opened 2024-10-17
Removed unused file

### LINKED ISSUE #3407: [Docs RFC] Clean up the redirects.
There are a billion and one redirects in https://github.com/timescale/docs/blob/latest/.data/redirects.js. Please remove redirects to non-existent pages,  redirects to redirects, redirects that do not work. I don't think there is a way to see which ones are not used externally. Maybe ask @robsoncoelho if there is a tool to test the redirects. 

To make it more fun, there is also https://github.com/timescale/web-documentation/blob/production-v3/lambda/redirects.js. Not sure if these redirects are live. Once again, Robson will know.

atovpeko:


atovpeko:
@robsoncoelho I know we have previously agreed this file can be removed, but just want to make sure it's OK with you.

robsoncoelho:
> @robsoncoelho I know we have previously agreed this file can be removed, but just want to make sure it's OK with you.

LGTM!

billy-the-fish (approved):

PR #3514, opened 2024-10-18
update: changelog for 10/18/2024

solugebefola:
# Description

# Links

# Writing help

# Review checklists

### DIFF
--- a/about/changelog.md
+++ b/about/changelog.md
@@ -8,6 +8,15 @@ keywords: [changelog, upgrades, updates, releases]
 
 All the latest features and updates to Timescale products.
 
+## 💡 Help with hypertables and faster notebooks
+<Label type="date">October 18, 2024</Label>
+
+### 🧙Hypertable creation wizard
+After creating a service, users can now create a hypertable directly in the Timescale Console by first creating a table, then converting it into a hypertable. This is possible through the in-console PopSQL SQL editors. All standard hypertable configuration options are supported, as is editing the table schema directly in the console to anything the user would like. 
+
+### 🍭 PopSQL Notebooks 
+The newest version of PopSQL Notebooks is now waaaay faster.  Why? We've incorporated the newly developed v3 of our query engine that currently powers the Timescale Cloud console SQL Editor.  Check out the difference in query response times.
+
 ## ✨ Production-Ready Low-Downtime Migrations, MySQL Import, Actions Tab, and Current Lock Contention Visibility in SQL Editor
 <Label type="date">October 10, 2024</Label>
 
--- a/about/changelog.md
+++ b/about/changelog.md
@@ -12,10 +12,10 @@ All the latest features and updates to Timescale products.
 <Label type="date">October 18, 2024</Label>
 
 ### 🧙Hypertable creation wizard
-After creating a service, users can now create a hypertable directly in the Timescale Console by first creating a table, then converting it into a hypertable. This is possible through the in-console PopSQL SQL editors. All standard hypertable configuration options are supported, as is editing the table schema directly in the console to anything the user would like. 
+After creating a service, users can now create a hypertable directly in Timescale Console by first creating a table, then converting it into a hypertable. This is possible using the in-console SQL editor. All standard hypertable configuration options are supported, along with any customization of the underlying table schema. 
 
 ### 🍭 PopSQL Notebooks 
-The newest version of PopSQL Notebooks is now waaaay faster.  Why? We've incorporated the newly developed v3 of our query engine that currently powers the Timescale Cloud console SQL Editor.  Check out the difference in query response times.
+The newest version of Data Mode Notebooks is now waaaay faster.  Why? We've incorporated the newly developed v3 of our query engine that currently powers Timescale Console's SQL Editor.  Check out the difference in query response times.
 
 ## ✨ Production-Ready Low-Downtime Migrations, MySQL Import, Actions Tab, and Current Lock Contention Visibility in SQL Editor
 <Label type="date">October 10, 2024</Label>


about/changelog.md
```diff
@@ -8,6 +8,15 @@ keywords: [changelog, upgrades, updates, releases]
 
 All the latest features and updates to Timescale products.
 
+## 💡 Help with hypertables and faster notebooks
+<Label type="date">October 18, 2024</Label>
+
+### 🧙Hypertable creation wizard
+After creating a service, users can now create a hypertable directly in the Timescale Console by first creating a table, then converting it into a hypertable. This is possible through the in-console PopSQL SQL editors. All standard hypertable configuration options are supported, as is editing the table schema directly in the console to anything the user would like.
```
billy-the-fish:
```suggestion
After creating a service, users can now create a hypertable directly in Timescale Console by first creating a table, then converting it into a hypertable. This is possible using SQL editor. All standard hypertable configuration options are supported, as is editing the table schema directly in the console to anything the user would like. 
```

about/changelog.md
```diff
@@ -8,6 +8,15 @@ keywords: [changelog, upgrades, updates, releases]
...
 All the latest features and updates to Timescale products.
 
+## 💡 Help with hypertables and faster notebooks
+<Label type="date">October 18, 2024</Label>
+
+### 🧙Hypertable creation wizard
+After creating a service, users can now create a hypertable directly in the Timescale Console by first creating a table, then converting it into a hypertable. This is possible through the in-console PopSQL SQL editors. All standard hypertable configuration options are supported, as is editing the table schema directly in the console to anything the user would like. 
+
+### 🍭 PopSQL Notebooks 
+The newest version of PopSQL Notebooks is now waaaay faster.  Why? We've incorporated the newly developed v3 of our query engine that currently powers the Timescale Cloud console SQL Editor.  Check out the difference in query response times.
```
billy-the-fish: Are you saying:
- PopSQL (external web tool) Notebooks are faster because they now use the same query engine as SQL Editor
- Notebooks in Data mode (PopSQL in Timescale Console) are faster because it now uses the same query engine as SQL Editor.

SQL Editor is in Timescale Console, not Timescale Cloud console

billy-the-fish (approved):

about/changelog.md
solugebefola: Yah, I was specifically referencing the SQL editor.  This is where that comes from https://iobeam.slack.com/archives/C06RPB8EVMY/p1728941525019489?thread_ts=1728936236.129429&cid=C06RPB8EVMY
> Sure! Notebooks were first built a few years ago now, using the popsql-server query engine (uses [blazer](https://github.com/ankane/blazer) for ruby). While this engine "works" for most queries, it has some big inefficiencies. The biggest problem is that it buffers the result data in its entirety in server memory before sending the response to the client. For very large result sets, this becomes apparent as a large latency before any results are seen in the app. It also makes it possible for a single query to crash the server (by using all of its memory).
We then went on to tackle these problems, building a v2 query engine in node.js that focused on streaming the results. We used our existing Websockets server, along with socket.io-stream to pipe the bytes from server to client. This made many queries feel much faster, as we can start to render the first few rows while the rest of the data is still streaming in. While this was a huge improvement over the status-quo, we still were not satisfied. We relied on 3rd party node.js libraries for the database drivers, many of which we found to be quite unstable. We ended up forking and patching several of these, to avoid server crashes. We also had problems with client-side data management. Very large result sets still bogged us down, leading to slowness and client-side crashes.
So next, we built a client-side results cache, powered by DuckDB. This moved the bulk of the data transfer, processing, and storage to a web worker, freeing up the UI thread to always remain snappy.
Then, 
[@Nathan Cochran](https://iobeam.slack.com/team/U06S8H0V94P)
 decided to write a V3 query engine in Golang (now called popsql-query). This is based on a much more robust db driver foundation from the Go community, which has practically entirely eliminated our stability problem. We also took this opportunity to take another pass at the data transfer protocol. We found that we can achieve drastically better throughput using Apache Arrow's streaming data format, on top of raw HTTP (ditching websockets).
With all this innovation to the query engine, we focused on PopSQL's primary query interface - query documents. We've acquired a fair bit of tech debt, with other features still using some combination of the older engines. This includes notebooks, which, until today, were still on that V1 engine.
Most recently, we built the QueryWidget as a way to bundle our best solution into one reusable package: it uses the latest popsql-query (v3) engine, the duckdb results cache, our (relatively recently) rewritten monaco editor, and our latest results table component (using Glide over ag-grid). This was built with notebooks in mind: it should be something we can use there in addition to timescale console (our first target).
So, today, I've wrapped up exactly that: I refactored the notebooks code to leverage the same QueryWidget component, bringing all its advantages with it. This also included revamping the "query previous results" feature to use DuckDB (instead of sql.js). The result is a substantial improvement in both performance and UX. We still have a healthy list of improvements we'd like to make to notebooks, but this feels like quite the revival of a feature that had grown neglected.
Dashboards are next

about/changelog.md
solugebefola: I'm actually not sure what Justin is referring to, Data mode or the standalone PopSQL.  I guess they are probably the same now.  I'll update
I thought we were referring to it as Timescale Cloud now?

about/changelog.md
solugebefola: Oops that one was actually written by Grant.  I'll update it though 😄

billy-the-fish: Ok, I've updated the suggestion, that one is good to go.

about/changelog.md
billy-the-fish: Timescale Cloud is the whole shebang. 
Timescale Console is the UI you use to administer Timescale Cloud and interact with data in the individual services.

PR #3636, opened 2024-12-02
chore: update the hypertable links.

### LINKED ISSUE #3628: [Feedback] Page: /api/latest/hypertable/create_hypertable_old/

No

None

### How could we improve the Timescale documentation site?

The [hypertable create documentation under use timescale](https://docs.timescale.com/use-timescale/latest/hypertables/create/) links to the old api for creating a hypertable in a big note, but I cant find on the page anywhere it links to the new documentation. 

The old documentation page then has no reference where to go for the new documentation. It took me 20 minutes of bumbling back and forth reading other pages to finally see the `create_hypertable` above the `create_hypertable (old interface)` in the sidebar.

The create hypertables example should link to the API docs somewhere (probably in the note that references the old variant so both can be found easily), and the old api docs and new api docs should cross link to make discovering the opposite variant easier.

billy-the-fish:


api/create_hypertable.md
deefdragon, line 1: I think describes is the correct word here. api documentation *describes* how something is expected to be interfaces with. tutorial documents then *use* a given api.

only a minor bike shed tho. it otherwise lgtm.

billy-the-fish: Updated.

atovpeko (approved):

PR #3639, opened 2024-12-03
chore: add a note about chunk size.

billy-the-fish:
Follow up on https://timescaledb.slack.com/archives/C4GT3N90X/p1732012772234309?thread_ts=1732001663.561169&cid=C4GT3N90X.

### DIFF
--- a/use-timescale/hypertables/about-hypertables.md
+++ b/use-timescale/hypertables/about-hypertables.md
@@ -60,21 +60,22 @@ the number of chunks you see when inspecting it.
 ### Best practices for time partitioning
 
 Chunk size affects insert and query performance. You want a chunk small enough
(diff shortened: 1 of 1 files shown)


jonatas (approved):
Very good! Should we mention anything about uncompressed vs compressed chunks and how it should be tweak for compression?

Because we also have the [timescaledb.compress_chunk_time_interval](https://docs.timescale.com/api/latest/compression/alter_table_compression/) that can be useful for helping to keep the chunk size in some direction.

@svenklemm, @RobAtticus any other learning would worth to add/replace here?

billy-the-fish:
@jonatas Let's revisit when the  compress_chunk_time_interval when the API is no longer experimental.

PR #3715, opened 2025-01-17
Fix configuration name

XbaoWu:
# Description
There is an error in the config that the enable _ chunk _ skipping feature on the document.

# Links
Not yet


billy-the-fish (approved):
Thanks @XbaoWu , LGTM.

PR #3719, opened 2025-01-17
Fix Upgrade Matrix

philkra:
Closed https://github.com/timescale/docs/issues/3714

PR #3720, opened 2025-01-17
Follow support matrix

philkra:
Follow up after https://github.com/timescale/docs/pull/3719 -- reverted to incorrect state showing support for PG12 in v2.15 and v2.14


billy-the-fish (approved):
LGTM

PR #3868, opened 2025-02-24
docs: add early access Typescript note

danstarns:
We released early access to our Typescript integration last week and would like to capture some traffic and early adopters before a more solidified documentation change.


atovpeko:
Closing this PR because it does not look relevant anymore. If it is, please open a new one.

PR #3939, opened 2025-03-17
Update merge chunks docs

erimatnor:
In 2.19 it will be possible to merge compressed chunks, so the API docs should be updated to reflect that.


erimatnor:
@billy-the-fish Not 100% sure how you'd like to do this change. I just changed the current formulations, but I am wondering if we should have some kind of "Since 2.19" flag for merging compressed chunks?

billy-the-fish (approved):
LGTM.

billy-the-fish:
@erimatnor the merge chunks on compressed data should be in the release notes and the changelog. If we try and put that all over the doc it will become more tag than doc :-).

PR #3975, opened 2025-03-28
chore: set default open accordions

coelhucas:
Depends on https://github.com/timescale/web-documentation/pull/846

`defaultOpen` on `page-index.js` will determine whether an accordion will be open by default:

image


billy-the-fish (approved):
LGTM

PR #3999, opened 2025-04-08
chore: adding logos to integrations main page

victormeirelests:
Addin logos to integrations index page.
image

[Task](https://timescale-pmm.monday.com/boards/4681403565/views/104841279/pulses/8873511579?userId=46253355)
[Figma](https://www.figma.com/design/fDVDWURwz7tooc8pR1oaj8/Miscellaneous?node-id=6739-51&p=f&t=FoJJHsoJDRyTCgrC-0)

### DIFF
--- a/use-timescale/integrations/index.md
+++ b/use-timescale/integrations/index.md
@@ -20,21 +20,21 @@ Some of the most in-demand integrations are listed below.
 
 |                Name                | Description                                                               |
 |:----------------------------------:|---------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


billy-the-fish (changes requested):
The images don't seem to be the right size, and are above the link rather than next to it.

victormeirelests:
> The images don't seem to be the right size, and are above the link rather than next to it.

Yes yes, still a work in progress

PR #4066, opened 2025-04-30
Update about high cardinality vs compression

### LINKED ISSUE #4056: [Docs RFC]Add information about cardinality to improve compression
The most common reason people have a low compression rate is that they choose a segment key with very high cardinality with respect to the chunk, so it actually doesn't group together many pre-compressed rows in each compressed batch

Add information in:

* https://docs.timescale.com/use-timescale/latest/compression/troubleshooting/
* https://docs.timescale.com/use-timescale/latest/hypercore/troubleshooting/
* https://docs.timescale.com/use-timescale/latest/hypercore/real-time-analytics-in-hypercore/

Use information in https://www.timescale.com/blog/what-is-high-cardinality to explain.

atovpeko:


thinkingbox (approved):
Very helpful, thanks!

billy-the-fish (approved):

PR #4067, opened 2025-04-30
Do not recommend setting Hypercore TAM on hypertable root

erimatnor:
Update the docs to recommend converting chunks to Hypercore TAM using the columnstore policy instead of enabling it on new chunks.

Using Hypercore TAM on new chunks is suboptimal because there will be no data compressed anyway until the policy runs, leading to a number of confusing issues, such as no updated columnstore stats.


billy-the-fish (approved):
Nice.

mkindahl (approved):

PR #4081, opened 2025-05-06
Align products and pricing plans

atovpeko:
NOTE: pull from latest and update for prometheus


billy-the-fish:
Changelog also is for self-hosted. 
Migration is not for MST

dannydanlev (approved):

PR #4115, opened 2025-05-14
Clarify the matter with foreign keys to/from hypertables

### LINKED ISSUE #3904: [Docs RFC]Revisit the hypertables docs about foreign keys
I feel like our docs are actually a huge mess for how this kind of stuff works.
When I google hypertable foreign keys, I find this one: https://docs.timescale.com/self-hosted/latest/distributed-hypertables/foreign-keys/
Which is about distributed hypertables which are apparently deprecated for a long time already. So not sure why this outperforms our more generic docs:
https://docs.timescale.com/use-timescale/latest/schema-management/about-constraints/
This docs page mentions:
Hypertables support all standard PostgreSQL constraint types, except for foreign key constraints from a hypertable referencing another hypertable.
Which is good but google ranks this closed and not fixed github issue result higher:
https://github.com/timescale/timescaledb/issues/1394
This links to: https://github.com/timescale/timescaledb/issues/498 which is apparently an open feature request for foreign keys in hypertables
However there is a comment in here that links to this PR: https://github.com/timescale/timescaledb/pull/6989
Which says:
Currrently we only allow Hypertables references other tables, with this patch the opposite direction is supported as well and tables can have foreign key references into hypertables.
So what now? Foreign keys are an open issue while they don't support foreign keys into other hypertables (or rather they do according to the PR?)
To make it even better when you look generically for hypertable limitations I find this:
https://docs.timescale.com/use-timescale/latest/limitations/
Which again mentions that foreign keys between hypertables don't work? And also talks a whole lot about a long time deprecated feature with distributed hypertables without the docs page being marked as deprecated.
My take away would be, if I can trust the PR it all works and all the open and closed issues as well as the docs are inconsistent and lying. (edited)

atovpeko:


billy-the-fish (approved):

atovpeko:
Approved by reporter via Slack

PR #4128, opened 2025-05-19
Updates to prometheus docs

atovpeko:


billy-the-fish (approved):

ramonguiu:
@atovpeko @rahilsondhi  the instructions for self-hosted related to adding the Postgres exporter also apply to TS Cloud. The TS Cloud set up to get infrastructure metrics (cpu, memory, storage) are in addition to the ones from the Postgres exporter. The thing we need to be careful of is making sure the Prometheus labels for the Postgres exporter are configured match the ones set by TS Cloud. 

I've tested the TS Cloud exporter for Prometheus and metrics look like this:
```
# TYPE timescale_cloud_system_cpu_total_millicores gauge
timescale_cloud_system_cpu_total_millicores{project_id="mzel21cd40",region="us-east-1",role="primary",service_id="p7s4vw4ik7",service_type="single-node"} 500
timescale_cloud_system_cpu_total_millicores{project_id="mzel21cd40",region="us-east-1",role="replica",service_id="p7s4vw4ik7",service_type="single-node"} 500
```
So we need to tell the user to configure the following labels in their Postgres exporter for each service:
- project_id
- service_id
- region

This way they will be able to easily create dashboards that can be filtered using those parameters.

atovpeko:
> @atovpeko @rahilsondhi the instructions for self-hosted related to adding the Postgres exporter also apply to TS Cloud. The TS Cloud set up to get infrastructure metrics (cpu, memory, storage) are in addition to the ones from the Postgres exporter. The thing we need to be careful of is making sure the Prometheus labels for the Postgres exporter are configured match the ones set by TS Cloud.
> 
> I've tested the TS Cloud exporter for Prometheus and metrics look like this:
> 
> ```
> # TYPE timescale_cloud_system_cpu_total_millicores gauge
> timescale_cloud_system_cpu_total_millicores{project_id="mzel21cd40",region="us-east-1",role="primary",service_id="p7s4vw4ik7",service_type="single-node"} 500
> timescale_cloud_system_cpu_total_millicores{project_id="mzel21cd40",region="us-east-1",role="replica",service_id="p7s4vw4ik7",service_type="single-node"} 500
> ```
> 
> So we need to tell the user to configure the following labels in their Postgres exporter for each service:
> 
> * project_id
> * service_id
> * region
> 
> This way they will be able to easily create dashboards that can be filtered using those parameters.

Here is a PR for this update: https://github.com/timescale/docs/pull/4137

PR #4217, opened 2025-06-19
chore: add links for the more detailed explanation to the hypercore docs

billy-the-fish:


brandonpurcell-dev (approved):
I had some issues with the page rendering in the build process so I couldn't preview it but the code looks good.

PR #4311, opened 2025-07-25
Add region to CLI

philkra:
Intentially define which region the S3 table bucket should be created in


use-timescale/tigerlake.md
```diff
@@ -81,6 +81,7 @@ To connect a $SERVICE_LONG to your data lake:
 
    Replace the following values in the command, then run it from the terminal:
 
+   * `Region`: region of the S3 table bucket
```
atovpeko, line 84:
```suggestion
   * `Region`: the region of the S3 table bucket
```

atovpeko (approved):

PR #4376, opened 2025-08-26
Add llms full

billy-the-fish:


robsoncoelho (approved):
I recommend we move the generate_llms_full.py file to web-documentation before merging to production.

billy-the-fish:
Hi @robsoncoelho, not sure what you think. I have added the generated file to https://github.com/timescale/web-documentation/pull/967 so you can do your magic. I kept the helper script in the .helper-scripts directory to match what was done for the GUCs generation. I imagine there has to be some kind of magic done to run this script when we build to production. 

If you prefer the direct move to web-documentation, I will do that now.

PR #4378, opened 2025-08-26
chore: update manual convert all chunks.

billy-the-fish:


antonum (approved):
New syntax looks really clumsy, comparing to legacy:
```
SELECT compress_chunk(c, true) FROM show_chunks('transactions') c;
```
Approving, but API review is suggested.

PR #4395, opened 2025-09-04
Add UUID API functions

erimatnor:
# Description

Add API documentation for the new UUID functions that were introduced in 2.22.

# Links

https://github.com/timescale/timescaledb/pull/8546

# Review checklists


erimatnor:
I am not sure I did things right and I probably missed a bunch of things. But at least it is a starting point.

api/uuid-functions/uuid_timestamp.md
```diff
@@ -0,0 +1,36 @@
...
+
+# uuid_timestamp() <Tag type="Community">Community</Tag>
+
+Extract a PostgreSQL timestamp with time zone from a version 7
+UUID. The UUID contains a millisecond unix timestamp and an optional
+sub-millisecond fraction. Only the millisecond part is used to
+construct the PostgreSQL timestamp.
+
+To include the optional sub-millisecond fraction in the returned
+timestamp, use `uuid_timestamp_micros`.
```
erimatnor: Should probably add link to other function here.

api/uuid-functions/uuid_timestamp_micros.md
```diff
@@ -0,0 +1,37 @@
...
+  license: community
+  type: function
+products: [cloud, mst, self_hosted]
+---
+
+# uuid_timestamp_micros() <Tag type="Community">Community</Tag>
+
+Extract a PostgreSQL timestamp with time zone from a version 7
+UUID. The UUID contains a millisecond unix timestamp and an optional
+sub-millisecond fraction. Unlike `uuid_timestamp`, the microsecond
```
erimatnor: Should probably add link to other function here.

billy-the-fish: Done

api/uuid-functions/uuid_timestamp.md
billy-the-fish: Done

api/hypertable/create_hypertable.md
```diff
@@ -121,6 +122,28 @@ CREATE FUNCTION event_started(jsonb)
 SELECT create_hypertable('events', by_range('event', partition_func => 'event_started'));
 ```
 
+### Time partition a hypertable using [UUIDv7][uuidv7_functions]:
+
+1. Create a table with a UUIDv7 column:
+    ```sql
+    CREATE TABLE events (
+        id  uuid PRIMARY KEY DEFAULT uuidv7(),
```
erimatnor, line 144:
```suggestion
        id  uuid PRIMARY KEY DEFAULT generate_uuidv7(),
```

api/hypertable/create_hypertable.md
```diff
@@ -121,6 +122,28 @@ CREATE FUNCTION event_started(jsonb)
...
+    ```
+
+1. Partition the table based on the timestamps embedded within the UUID values:
+
+    ```sql
+    SELECT create_hypertable(
+        'events',
+        by_uuidv7('id'),
+        chunk_time_interval => INTERVAL '1 month'
+    );
```
erimatnor:
```suggestion
   SELECT create_hypertable(
        'events',                                      
        by_range('id', INTERVAL '1 month')
    );
```

api/hypertable/create_table.md
```diff
@@ -79,6 +79,16 @@ arguments specific to $TIMESCALE_DB.
...
    ```
 
+- **Create a $HYPERTABLE partitioned using [UUIDv7][uuidv7_functions]**:
+
+   ```sql
+    CREATE TABLE events (
+      vvid  uuid PRIMARY KEY DEFAULT uuidv7(),
+      vvpayload jsonb
+    );
+    WITH (tsdb.hypertable, tsdb.partition_column = by_uuidv7('id'));
```
erimatnor:
```suggestion
   ```sql
   CREATE TABLE events (
      id  uuid PRIMARY KEY DEFAULT generate_uuidv7(),
      payload jsonb
    ) WITH (tsdb.hypertable, tsdb.partition_column = 'id');
```

erimatnor: AFAIK, this format of create table automatically enables compression as well. I believe (optimized) UUID compression is not enabled by default, which means that the `id` column will use sub-optimal compression. I am wondering if we should preface this with saying that we recommend setting `enable_uuid_compression=true` first?

erimatnor: @dbeck Do you have any insights here?

api/uuid-functions/index.md
```diff
@@ -0,0 +1,83 @@
...
+keywords: [jobs]
+tags: [background jobs, scheduled jobs, automation framework]
+products: [cloud, mst, self_hosted]
+---
+
+# UUIDv7 functions <Tag type="community">Community</Tag>
+
+
+UUIDv7 is a time-ordered UUID that includes a Unix timestamp (with millisecond precision) in its first 48 bits. Like 
+other UUIDs, it uses 6 bits for version and variant info, and the remaining 74 bits are random.
```
erimatnor, line 13: The precision thing can be a bit confusing, so I think we need to be very clear on it.  Here we write millisecond precision in the text but show a picture with microseconds, which can raise questions.

The UUIDv7 specifies a millisecond timestamp. However, one can add an optional sub-millisecond fraction, which we do in this function.

Therefore the generated UUID will have microsecond precision. The caveat, though, is that there is no way to specify in the UUID that this extra precision is encoded in the extra bits so if someone gets this UUID and doesn't know that the extra bits encode microseconds, there is no way to know for sure that the precision is microseconds.

Not sure how we can make this clear without going into too much detail.

Maybe the picture should say that the microsecond bits are optional?

api/hypertable/create_hypertable.md
erimatnor: Probably this example was taken from an outdated source?

erimatnor: FWIW, the PG18-equivalent function will be called `uuidv7()`, but I don't think we want to rely on it because it doesn't exist in older PostgreSQL versions.

Can we have different docs for pg18 vs older versions of PostgreSQL so that this example uses different functions depending on version?

erimatnor (commented):
Did a quick pass to catch the most obvious things. Will do a more thorough pass later.

api/hypertable/create_hypertable.md
billy-the-fish: Updated with tabs.

api/hypertable/create_table.md
billy-the-fish: Updated with guc setting and tabs for the postgres versions

api/uuid-functions/index.md
billy-the-fish: I've updated the image. Will take a while for the AWS cache to clear to make it visible. Added the updated image here for speed. 
uuidv7-structure-microseconds

api/uuid-functions/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+- **Workflow / durable execution runs**: 
+
+   Each run needs a stable ID for joins and retries, and you often ask “what started since X?”. UUIDs help by serving
+   both as the primary key and a time cursor across services. For example:
+
+    ```sql
+    SELECT run_id, status
+    FROM runs
+    WHERE to_timestamp(run_id) >= now() - interval '5 minutes';
```
brandonpurcell-dev: In our examples it would be good to differentiate between this approach that has to scan and apply the to_timestamp() function to every row vs using to_uuidv7_boundary() that should be more efficient since it's limiting by a range with the random portion zeroed out.

billy-the-fish: Sounds good. Could you supply a code example please.

billy-the-fish (approved):

api/uuid-functions/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+- **High-rate event logs for observability and metrics**: 
+
+   UUIDv7 gives you globally unique IDs (for traceability) and time windows (“last hour”), without the need for a 
+   separate `created_at` column. UUIDv7 create less churn because inserts land at the end of the index, and you can 
+   filter by time using UUIDv7 objects.
+
+  - Last hour:
+      ```sql
+      SELECT count(*) FROM logs WHERE to_timestamp(id) >= now() - interval '1 hour';
```
brandonpurcell-dev: We should use the boundary functions in our examples in the where clause vs to_timestamp(). In my testing it's up to 40x faster to use the boundary functions since it has to scan the entire column with to_timestamp() and it does not skip chunks as well. 

SELECT count(*) FROM logs WHERE id >= to_uuidv7_boundary(now() - interval '1 hour')

api/uuid-functions/index.md
brandonpurcell-dev: SELECT run_id, status
FROM runs
WHERE run_id >= to_uuidv7_boundary(now() - interval '5 minutes')

brandonpurcell-dev (commented):
Until we have another way to filter by time with UUIDv7 I recommend we only use the boundary functions in the where clause for performance reasons. I added examples. I've found that it's 40x faster as the datasets grow and uses a lot less memory.

brandonpurcell-dev (approved):
Looks good, hoping to get 2.22.1 out soon so we can get users leveraging this.

PR #4407, opened 2025-09-09
Mark chunk skipping as experimental

philkra:
The feature is not beta or GA, and this PR marks the proper stability level of experimental.


billy-the-fish:
i thought we were moving away from experimental to early access

philkra:
No objection to mark it as EA. Can you please share what partials are there to mark it as such?

billy-the-fish (approved):

PR #4420, opened 2025-09-12
chore: add note about chunks.

### LINKED ISSUE #4419: [Docs RFC]Add a note about chunk intervals

billy-the-fish:


jkoria (approved):

api/hypertable/set_chunk_time_interval.md
```diff
@@ -32,12 +32,20 @@ UNIX epoch, set `chunk_time_interval` to 24 hours:
...
+|`hypertable`|REGCLASS| -       | ✔                                                                    | Hypertable or continuous aggregate to update interval for.                                                                                       |
+|`chunk_time_interval`|See note|-       | ✔   | Event time that each new chunk covers.                                                                                                           |
+|`dimension_name`|REGCLASS|-       | ✖ | The name of the time dimension to set the number of partitions for. Only use `dimension_name` when your hypertable has multiple time dimensions. |
+
+If you change chunk time interval from 7-days to 3-days, you may have a remainder in the transition chunk: 1 day + 3-days + 3-days = 7.
+This happens because the `chunk_time_interval` divides the timeline starting from epoch 0. For example, two 7-day 
+chunks cover 14 days. Switching to 3-day chunks aligns with a 15-day cycle, leaving a remainder 1-day chunk.
```
erimatnor: Here's a stab at trying to explain in more detail. Maybe it is too much? Use as you wish or don't use at all.

```suggestion
If you change chunk time interval it might result in a chunk that is smaller than the new interval. For example, if you have two 7-day chunks that cover 14 days and change the `chunk_time_interval` to 3 days, you may end up with a transition chunk covering only one day. This happens because the start and end of the new chunk is calculated based on dividing the timeline by the `chunk_time_interval` starting at epoch 0, leading to chunks [0, 3), [3, 6), [6, 9), [9, 12), [12, 15), [15, 18) and so on. The two 7-day chunks already covered data up to day 14, so the 3-day chunk for [12, 15) will be cut down to a one day chunk. The following chunk at [15, 18) will be created as a full 3 day chunk.
```

erimatnor (approved):

PR #4460, opened 2025-09-30
Refresh data tiering

atovpeko:


use-timescale/data-tiering/index.md
```diff
@@ -10,11 +10,14 @@ tags: [storage, data management]
 # Tiered storage
 
 Tiered storage is a [hierarchical storage management architecture][hierarchical-storage] for 
-[time series and analytics][create-service] $SERVICE_SHORT you create in [$CLOUD_LONG](https://console.cloud.timescale.com/).
+[real-time analytics][create-service] $SERVICE_SHORTs you create in [$CLOUD_LONG](https://console.cloud.timescale.com/).
```
billy-the-fish, line 13: Nice, but a bit too short. Can you add something like the following before the image. :

You use the high-performance tier for data that requires quick access, and the object tier for rarely used historical data. Tiering policies move older data asynchronously and periodically from high-performance to low-cost storage, sparing you the need to do it manually. Chunks from a single hypertable, including compressed chunks, can stretch across the storage tiers.

billy-the-fish (approved):
Small nit.

use-timescale/data-tiering/index.md
atovpeko: Up for discussion? This just repeats what is said a bit lower in the page, and that page is not big enough to justify this summary.

atovpeko: We could also link to this page instead: https://docs.tigerdata.com/use-timescale/latest/data-tiering/about-data-tiering/ (I can move the diagram here, too)

PR #4472, opened 2025-10-02
chore: use sparse_index explicitly for, sparse indexes :-).

billy-the-fish:


cevian (approved):

Poroma-Banerjee:
FYI. all these are valid aliases, "compress_index", "compress_sparse_index", "index", "sparse_index"

Poroma-Banerjee (approved):

PR #4476, opened 2025-10-03
Rework pricing with tabs

atovpeko:


atovpeko:
- To review: https://docs-dev.timescale.com/docs-405-rework-pricing-with-tabs/about/405-rework-pricing-with-tabs/pricing-and-account-management/
- To confirm/clarify: instance sizes for Azure (same as AWS?) If different, also update https://docs.tigerdata.com/use-timescale/latest/services/change-resources/

minkimipt (approved):
I don't see any changes related to Azure here. The content looks correct to me. But I'll leave other checks to @billy-the-fish

atovpeko:
> I don't see any changes related to Azure here. The content looks correct to me. But I'll leave other checks to @billy-the-fish

https://docs-dev.timescale.com/docs-405-rework-pricing-with-tabs/about/405-rework-pricing-with-tabs/pricing-and-account-management/ now has two tabs, the one for Azure does not mention data tiering and other features that are not available in Azure

_partials/_billing-example.md
```diff
@@ -0,0 +1,20 @@
...
+- **Storage**: metered based on your average GB consumption per hour. Storage grows and shrinks automatically
+  with your data.
+
+Your monthly price for compute and storage is computed similarly. For example, over the last month your
+$SERVICE_LONG has been running compute for 500 hours total:
+- 375 hours with 2 CPU
+- 125 hours 4 CPU
+
+**Compute cost** = (`375` x `hourly price for 2 CPU`) + (`125` x `hourly price for 4 CPU`)
+
```
billy-the-fish, line 16: Any chance you could add an example for storage as well please. 

Plus, can you add something saying that these charges run even when you are not connected.  And to reduce costs, pause the service. The same line should be in Monitor usage and costs.

atovpeko: will add in the free tier scope, as discussed

PR #4557, opened 2025-11-07
invalidation engine clarification for CAGGs

### LINKED ISSUE #4552: [Feedback]

None

None

### How could we improve the Tiger Data documentation site?

The docs for continuous aggregates ([[About continuous aggregates](https://docs.tigerdata.com/use-timescale/latest/continuous-aggregates/about-continuous-aggregates/#about-continuous-aggregates)](https://docs.tigerdata.com/use-timescale/latest/continuous-aggregates/about-continuous-aggregates/)) state the following

> When data older than the invalidation threshold is changed, the maximum and minimum timestamps of the changed rows is logged, and the values are used to determine which rows in the aggregation table need to be recalculated. This logging does cause some write load, but because the threshold lags behind the area of data that is currently changing, the writes are small and rare.

From my understanding (and talking to TigerData engineers) this statement is not entirely valid? Or is outdated? Only the buckets of continuous aggregate that are changed by modified rows are recalculated but not the entire range of bucket as indicated by the min/max note.

atovpeko:


billy-the-fish (approved):

use-timescale/continuous-aggregates/about-continuous-aggregates.md
```diff
@@ -260,11 +260,17 @@ materialization threshold forward in time. This ensures that the threshold lags
...
+operation logs the minimum and maximum timestamps of the rows it modified.
+The continuous aggregate then identifies which complete time buckets contain
+these modified rows. Only those specific buckets are recalculated during refresh.
+
+For example, if one operation modifies rows in the 10:00 bucket and another
+modifies rows in the 15:00 bucket, only those two buckets are recalculated. The
+buckets in between (11:00, 12:00, 13:00, 14:00) are not affected.
+
+This logging does cause some write load. However, the threshold lags behind the
+area of data that is currently changing, so the writes are small and rare.
```
svenklemm: Hmm its more complicated than that, it depends how you define operation, the min/max tracking is per transaction so if you modify 10:00 and 15:00 in 1 transaction then all buckets in between would also be recalculated. If you modify 10:00 in a transaction and 15:00 in a separate transaction then intermediate values would not be recalculated

atovpeko: Updated, thank you.

svenklemm (approved):

PR #4588, opened 2025-11-21
Update to secondary indexes doc

### LINKED ISSUE #4187: [Feedback] Page: /use-timescale/latest/hypercore/secondary-indexes/

None

No

### How could we improve the Timescale documentation site?

The query performance comparison between _Execute a query on a regular hypertable_ and _Execute a query on the same data segmented and ordered in the columnstore_ is not only unfair but also not correct:

1. Why doesn't the first query on the regular hypertable use a `time` filter? The query planning is always less performant because of this reported issue https://github.com/timescale/timescaledb/issues/7646
2. The output of the second query is not correct obviously - the output has `device_id`  and `avg_disk_io` columns which are not in the query

atovpeko:


use-timescale/hypercore/secondary-indexes.md
```diff
@@ -60,44 +60,51 @@ at the right time. When you segment your data to access specific columns, your q
 For example, to access information about a single device with a specific `device_id`, you segment on the `device_id` column. 
 This enables you to run analytical queries on compressed data in the $COLUMNSTORE much faster.
 
-For example for the following $HYPERTABLE:
+To illustrate, let's create a $HYPERTABLE and then run the same query on it with and without optimizations:
```
billy-the-fish: I don't think this will work. Columnstore is enabled by default from 2.23 on.

billy-the-fish:
```suggestion
To illustrate, run the same query on a $HYPERTABLE, first without, then with optimizations:
```

use-timescale/hypercore/secondary-indexes.md
```diff
@@ -60,44 +60,51 @@ at the right time. When you segment your data to access specific columns, your q
...
 
 1. **Execute a query on the same data segmented and ordered in the $COLUMNSTORE**
 
-   1. Control the way your data is ordered in the $COLUMNSTORE:
-  
+   1. Control the way your data is ordered and segmented in the $COLUMNSTORE:
+
       ```sql
       ALTER TABLE metrics SET (
         timescaledb.enable_columnstore = true,
```
billy-the-fish: Columnstore is already enabled, you don't need to enable it.

billy-the-fish (commented):
One nit, and can you check the query without enable_columnstore please.

billy-the-fish (approved):
Nice.

PR #4655, opened 2026-02-04
Update for max_connections config

atovpeko:


atovpeko:
I hope this makes sense based on what we have in the UI. As I understand 1GB service from your table is some edge case that shouldn't really be documented?

minkimipt (approved):
Thank you for your changes. The document accurately reflects the code.

PR #4658, opened 2026-02-06
Update for tiering on azure

atovpeko:
Removed notes about tiered storage not available for Azure, so there are now generic pages about tiering: https://docs-dev.timescale.com/docs-azure-data-tiering/use-timescale/azure-data-tiering/data-tiering/ The only note is left for enhanced storage, which is not available on Azure.


_partials/_disaggregated-compute-storage-azure.md
```diff
@@ -8,10 +8,13 @@ provisioning your $SERVICE_SHORTs or later, as your needs grow.
   the same performance. The best way to size your needs is to sign up for a free trial and test
   with a realistic workload.
 
-* **Storage**: pay only for the storage you consume. The high-performance storage offers you up to 64 TB of compressed
-  (typically 80-100 TB uncompressed) data and is metered on your average GB consumption per hour. We can help you compress your data by up to 98% so you pay even less. 
+* **Storage**: pay only for the storage you consume. You have high-performance storage for more-accessed data, and
```
steve-tsdb: recent or frequently accessed data

_partials/_disaggregated-compute-storage-azure.md
```diff
@@ -8,10 +8,13 @@ provisioning your $SERVICE_SHORTs or later, as your needs grow.
...
+  [low-cost bottomless storage in S3][data-tiering] for other data. The high-performance storage offers you up to 64 TB of compressed
```
steve-tsdb: older or less frequently accessed data.

about/feature-comparison.md
```diff
@@ -44,8 +44,8 @@ The following table compares the features available in $CLOUD_LONG and self-host
...
 | Dynamic I/O provisioning for high-read/ high-write performance                                                      | <span style="color:green">✓</span>  | <span style="color:green">✓</span>  | Manual                                      |
-| Low-cost storage with infinite capacity on S3                                                                       | <span style="color:green">✓</span>  | <span style="color:red">✗</span>    | <span style="color:red">✗</span>            |
-| Transparent queries across high-performance and low-cost tiers                                                      | <span style="color:green">✓</span>  | <span style="color:red">✗</span>    | <span style="color:red">✗</span>            |
+| Low-cost storage with infinite capacity on S3                                                                       | <span style="color:green">✓</span>  | <span style="color:green">✓</span>  | <span style="color:red">✗</span>            |
```
steve-tsdb, line 47: on cloud object storage (rather than S3)

use-timescale/data-tiering/about-data-tiering.md
```diff
@@ -20,14 +20,14 @@ The tiered storage architecture in $CLOUD_LONG includes a high-performance stora
...
+High-performance storage is where your data is stored by default, until you [enable tiered storage][low-cost-storage] and [move older data to the low-cost tier][creating-data-tiering-policy]. In the high-performance storage, your data is stored in the block format and optimized for frequent querying. The [$HYPERCORE row-columnar storage engine][hypercore] enables you to compress the data in the high-performance storage by up to 98%, while improving performance. Coupled with other optimizations, $CLOUD_LONG high-performance storage makes sure your data is always accessible and your queries run at lightning speed.
```
steve-tsdb: ... your data is stored on a block device that is optimized for frequent querying...

steve-tsdb: "always accessible and your queries run at lightning speed" : )

I realize this is the current copy, but maybe something less hand-wavy and more accurate: "... e.g., "[your data] is durable and available and your queries run as quickly as they can"

steve-tsdb (commented):
couple minor comments/suggestions

about/feature-comparison.md
atovpeko: Do we want to remove all mentions of S3 then? If so, this has to be coordinated with marketing as well, because it's definitely on the pricing page: https://www.tigerdata.com/pricing (and across other places in docs)

use-timescale/data-tiering/about-data-tiering.md
atovpeko: Ah yes, the former attempts to put marketing into docs :-) Updated.

about/feature-comparison.md
philkra: Technically, Azure does not call "it" S3, but blob storage. That being said, S3 is an industry-wide term, LGTM to go with S3

philkra (approved):
LGTM

atovpeko:
@steve-tsdb I have realized there are UI updates for setting up tiering policies in Console. Do we need those steps described on https://www.tigerdata.com/docs/use-timescale/latest/data-tiering/enabling-data-tiering#low-cost-object-storage-tier or will it suffice to have SQL commands there (as it is now)?

about/feature-comparison.md
steve-tsdb: I don't want to create more work, but I think the problem here is that can be read _literally_ i.e. that we're tiering your data from your azure service on aws s3.

steve-tsdb (approved):
Still have an open question on the use of "S3" but can be async. No issues with using the commands.

